PR #3, opened 2024-12-19
Make the client log (most things) to debug or warn instead of info

svanharmelen:


einarmo (approved):
All this is reasonable.

PR #5, opened 2024-12-19
Export SubscriptionActivity for use when driving the session event-loop

svanharmelen:


einarmo (approved):
There's probably going to be a few of these, apologies in advance.

PR #7, opened 2024-12-19
Avoid implementation.rs

einarmo:
Use more descriptive names or move into `mod.rs` where appropriate.

I'm open for other suggestions.


opcua-client/src/browser/run.rs
svanharmelen, line 1: Or `browse.rs` which is also different from `brwoser` so will prevent the clippy warning?

svanharmelen (commented):
One suggestion, as for the rest; this is normally what I do as well... Use the mod file for both some (re)pub and include stuff as well as some main/generic module code 👍🏻

svanharmelen:
Nice one 👍🏻 

I did noticed that you merged `master` into the branch before doing a `squash and merge` both here and in some of the other PR's. Is there a specific reason for using this approach?

A `rebase` on the feature branch usually results in a much cleaner git history. Which actually touches the other one, using `squash and merge` instead of using a `merge commit`. While a squash of course "cleans" everything up by squashing everything into a single commit (so in a way that solves the effects of merging master into your feature branch), it does remove some history or separation that someone might have added on purpose.

I myself, for example, always make sure my PR's consist of one commit unless there are several related changed that touch different part of the code base. Having separate commits with a reasonable description in those cases helps others to see what was done and why. But you loose that info when you squash everything into a single commit which is why I personally never use that option.

And of course if people then offer a PR that isn't already cleanup up, I need to ask them to do so before I will merge it. But I still prefer that over loosing information about the code changes that are added.

oroulet:
Only one commit per MR sounds a bit too rigid (although that is also my personal goal).
But yes rebasing is much cleaner. Those merge commits inside a branch are only noise

svanharmelen:
I didn't say only one commit should be the norm 😏

Of course it depends on what you are adding and if making separate commits improves understanding and readability of the changes.

But either way squashing it all into one commit on merge will make you loose that info, which was my main point 🙂

einarmo:
I have two ways I usually do PRs, and that I'd prefer for this repo:

 - Mostly ignore the git history inside the PR, then squash everything when merging. This is good for bugfixes and other changes with a reasonably small diff, i.e. where a single commit is a good end-state.
 - Rebase and create a clean git history before rebasing onto master. This is good for larger changes, but takes a bit more work. I like bisectable commits, but there is no way to enforce that in github AFAIK.

For this PR I think a single commit is fine, so I don't give much thought to the history in the PR, and squash everything at the end.

If I haven't already I'm disabling merge commits to master, so you can either rebase or squash.

einarmo:
I'll accept some criticism that I'm sometimes a bit lazy about creating a clean git history inside a PR. It is the case in Rust that it often takes a lot of effort to get nicely bisectable commits, especially for more complicated features. I dislike having commits on master that don't build.

svanharmelen:
Hmm... That's a shame... But I guess it's your fork, so what can I say about it 🤷🏻‍♂️

einarmo:
I don't really understand your reaction, I think I'm mostly agreeing with you, just with the additional requirement that the commit history should be bisectable. Functionally, there's no real difference to squashing your branch into a single commit, then rebasing that on master, and using the "Squash and merge" feature in github. I'm saying I think that's an acceptable option if the PR ends up being small.

In the end, what matters is:

 - Is the PR reviewable?
 - What does the final history when applied to master look like?

I prefer to avoid merge commits on master, since they can make the history very messy, which means either using `rebase` and manually maintaining a clean history on the feature branch, or, for smaller patches, using "squash and merge".

svanharmelen:
Got it... Guess I (partly) misunderstood your previous reply 😏

One thing that's still not really clear to me is how merge commits mess up your git history?

It was (is?) the defacto standard for a long time and most (if not all) repo's I work with (both private and OSS) still use it without any issues.

But I guess that is a less important detail and maybe just a choice without real impact on the git history itself.

Anyway, sorry I kind of misread your initial response and glad we're actually more or less on the same page 😊

einarmo:
I think merge commits can be fine, but in that case they should be used exclusively IMO. Since we require up-to-date branches I suppose they don't mess up the history that much. I personally am still not a huge fan because I really like being able to look at the git history and immediately understand what is going on. Merge commits do not tend to have useful titles or descriptions, and so they are visual noise when trying to parse a git commit history.

For open source I also find the squash and merge feature to be really handy, since it's fairly common to get contributions from people with poor ability with git. Since these contributions will often be on the smaller side, squashing to a single commit is convenient. A minor advantage, I'll admit, but it's nice not to _always_ have to ask contributors to clean up their commit history. Of course, we will still have to do that for larger contributions, but that's fine.

Beyond that I think it's mostly just personal preference.

svanharmelen:
> I think merge commits can be fine, but in that case they should be used exclusively IMO

Yeah I agree with that thought...

> Merge commits do not tend to have useful titles or descriptions, and so they are visual noise when trying to parse a git commit history

Somewhat agree 😉 (in most cases people either use meaningful descriptions for both or neither).

> For open source I also find the squash and merge feature to be really handy, since it's fairly common to get contributions from people with poor ability with git. Since these contributions will often be on the smaller side, squashing to a single commit is convenient. A minor advantage, I'll admit, but it's nice not to always have to ask contributors to clean up their commit history.

While indeed a minor advantage, I guess its a fair point.

> Beyond that I think it's mostly just personal preference.

Definitely 100% agree on that one 😏 Thanks for taking the time to chat and I guess only squashing small things and rebasing the rest sounds like a decent "guideline" to me.

PR #15, opened 2025-01-01
Custom structures client

oroulet:


samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+        .await?
+        .into_iter()
+        .next()
+        .unwrap();
+    //dbg!(&res);
+    let Some(Variant::ExtensionObject(val)) = res.value else {
+        panic!("Unexpected variant type");
+    };
+    //dbg!(&val);
+    //let val: DynamicStructure = *val.into_inner_as().unwrap();
```
oroulet, line 137: that one works fine

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+    //dbg!(val.values());
+    let val: ErrorData = *val.into_inner_as().unwrap();
```
oroulet, line 139: is that possible at all? how? @einarmo ?

einarmo: This is just taking the value stored in the extension object and downcasting it to the type it really is. Getting it as `ErrorData` would require somehow deserializing from a `DynamicStructure` to a specific struct. I suppose something like that _could_ be implemented, but not trivially.

`into_inner_as` will only return `Some` if the type stored inside the extension object is the type you're calling it with. It won't do any clever conversions.

einarmo: If you want to actually deserialize to an `ErrorData` you need to create a type loader for that type, like you did on the server side. `DynamicStructure` is primarily intended to be used when you _can't_ do that. For example if you are creating a generic client like UAExpert, and you want to support custom structures, you need some way to deserialize from OPC-UA binary using only runtime information.

oroulet:
@einarmo ready now I think.

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        .await
+        .unwrap();
+    let handle = event_loop.spawn();
+    session.wait_for_connection().await;
+
+    let ns = session
+        .get_namespace_index(NAMESPACE_URI)
+        .await
+        .map_err(|e| format!("Error getting namespace index {:?}", e))?;
+    read_structure_var(&session, ns).await?;
```
svanharmelen, line 94:
```suggestion

    read_structure_var(&session, ns).await?;
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+                args.url.as_ref(),
+                SecurityPolicy::None.to_str(),
+                MessageSecurityMode::None,
+                UserTokenPolicy::anonymous(),
+            ),
+            IdentityToken::Anonymous,
+        )
+        .await
+        .unwrap();
+    let handle = event_loop.spawn();
```
svanharmelen, line 87:
```suggestion

    let handle = event_loop.spawn();
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    };
+    let val: DynamicStructure = *val.into_inner_as().unwrap();
+
+    dbg!(&val.get_field(0));
+    dbg!(&val.get_field(1));
+    dbg!(&val.get_field(2));
+    Ok(())
+}
+
+// the struct and enum code after that line could/should be shared with demo server
```
svanharmelen, line 138:
```suggestion
// The struct and enum code after this line could/should be shared with demo server,
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+// but having it here make the example selv contained
```
svanharmelen, line 139:
```suggestion
// but having it here makes the example self-contained.
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    Debug,
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    opcua::types::UaEnum,
+    opcua::types::BinaryEncodable,
+    opcua::types::BinaryDecodable,
+)]
+//#[cfg_attr(
```
svanharmelen, line 151:
```suggestion

//#[cfg_attr(
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[default]
+    Disabled = 1i32,
+    Enabled = 2i32,
+    Idle = 3i32,
+    MoveAbs = 4i32,
+    Error = 5i32,
+}
+
+#[derive(Debug, Clone, PartialEq, opcua::types::BinaryEncodable, opcua::types::BinaryDecodable)]
+//#[cfg_attr(
```
svanharmelen, line 168:
```suggestion

//#[cfg_attr(
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+//feature = "json",
+//derive(opcua::types::JsonEncodable, opcua::types::JsonDecodable)
+//)]
+//#[cfg_attr(feature = "xml", derive(opcua::types::FromXml))]
```
svanharmelen, line 172: Should these commented out lines stay in here?

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    Ok(())
+}
+
+async fn read_structure_var(session: &Arc<Session>, ns: u16) -> Result<(), StatusCode> {
+    let type_tree = DataTypeTreeBuilder::new(|f| f.namespace <= ns)
+        .build(session)
+        .await
+        .unwrap();
+    let type_tree = Arc::new(type_tree);
+    let loader = Arc::new(DynamicTypeLoader::new(type_tree.clone())) as Arc<dyn TypeLoader>;
```
svanharmelen, line 107:
```suggestion

    let loader = Arc::new(DynamicTypeLoader::new(type_tree.clone())) as Arc<dyn TypeLoader>;
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    session.add_type_loader(loader.clone());
+
+    let res = session
+        .translate_browse_paths_to_node_ids(&[BrowsePath {
+            starting_node: ObjectId::ObjectsFolder.into(),
+            relative_path: (&["ErrorData".into()][..]).into(),
+        }])
+        .await?;
+    let Some(target) = &res[0].targets else {
```
svanharmelen, line 116:
```suggestion

    let Some(target) = &res[0].targets else {
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        panic!("translate browse path did not return a NodeId")
+    };
+    let node_id = &target[0].target_id.node_id;
```
svanharmelen, line 119:
```suggestion

    let node_id = &target[0].target_id.node_id;
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    let dv = session
+        .read(&[node_id.into()], TimestampsToReturn::Neither, 0.0)
+        .await?
+        .into_iter()
+        .next()
+        .unwrap();
+    dbg!(&dv);
+    let Some(Variant::ExtensionObject(val)) = dv.value else {
```
svanharmelen, line 127:
```suggestion

    let Some(Variant::ExtensionObject(val)) = dv.value else {
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+        panic!("Unexpected variant type");
+    };
+    let val: DynamicStructure = *val.into_inner_as().unwrap();
```
svanharmelen, line 130:
```suggestion

    let val: DynamicStructure = *val.into_inner_as().unwrap();
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+
```
svanharmelen, line 131:
```suggestion
```

samples/custom-structures-client/src/main.rs
```diff
@@ -0,0 +1,178 @@
...
+    dbg!(&val.get_field(0));
+    dbg!(&val.get_field(1));
+    dbg!(&val.get_field(2));
+    Ok(())
```
svanharmelen, line 135:
```suggestion

    Ok(())
```

svanharmelen (dismissed):
Al lot of nitpicking, but I think good readable code is important (especially in an example, but in general as well).

opcua-types/src/impls.rs
```diff
@@ -414,3 +414,20 @@ impl Default for IdentityCriteriaType {
         Self::Anonymous
     }
 }
+
+impl From<&[QualifiedName]> for RelativePath {
+    fn from(value: &[QualifiedName]) -> Self {
+        let elements = value
+            .iter()
+            .map(|qn| super::relative_path_element::RelativePathElement {
```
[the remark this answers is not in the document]
oroulet: endret.

svanharmelen:
Noticed the changes (looks much better already), but there are still an number (4 or 5) of unresolved comments. Could you check those as well? Thanks!

oroulet:
OK I see that maybe I should show how to do automatic serialization to rust objects too or remove code related to that here. I jsut need to find out how to do that...

samples/custom-structures-client/src/main.rs
oroulet: @einarmo I tried to look around in code but could not find how to deserialize to ErrorData. Any pointer?

oroulet: @einarmo If you know something I would really appreciate. I really feel I am overlooking something not that hard here that should should be documented

einarmo: Sorry. You have two real options here.

First, you can do like you do on the server and create a `TypeLoader` that supports `ErrorData` instead of using `DynamicStructure`.

Your other option is to do something like this:

```rust
ErrorData {
    message: dyn_structure.get_value_by_name("Message").unwrap().clone().try_cast_to().unwrap(),
    error_id: dyn_structure.get_value_by_name("ErrorId").unwrap().clone().try_cast_to().unwrap(),
    ...
}
```

In theory we could make a macro to implement that automatically, and maybe have a way to consume a `DynamicStructure` and return its values so we can drop the clone.


Part of your problem here is that you are really meant to use only one of the two methods:

 - Either you know the types you need at compile time, so you make a `TypeLoader` to convert directly to those.
 - Or you don't, so you use `DynamicStructure`.

You are kinda trying to do both at the same time.

samples/custom-structures-client/src/bin/native_client.rs
```diff
@@ -0,0 +1,89 @@
...
+    handle.await.unwrap();
+    Ok(())
+}
+
+async fn read_structure_var(session: &Arc<Session>, ns: u16) -> Result<(), OpcUaError> {
+    let type_tree = DataTypeTreeBuilder::new(|f| f.namespace <= ns)
+        .build(session)
+        .await
+        .unwrap();
+    let type_tree = Arc::new(type_tree);
```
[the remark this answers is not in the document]
einarmo: Ahh... Right, to _decode_ the type you need a `TypeLoader`, you really should have that on the server as well.

Again, this isn't particularly easy to do by hand, and can probably be improved, you can use a static `TypeLoaderInstance`. See [async-opcua-types/src/generated/types/mod.rs](the generated code) for an example.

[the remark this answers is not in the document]
einarmo: Or maybe a better example is here: https://github.com/FreeOpcUa/rust-opcua/blob/master/samples/custom-codegen/src/generated/types/mod.rs

samples/custom-structures-client/src/main.rs
oroulet: thanks. made decoding works now.
Next is to look at encoding
but all these hardcoded node ids is not that great. need a closer look at it later

async-opcua-server/src/node_manager/memory/simple.rs
```diff
@@ -363,19 +363,30 @@ impl SimpleNodeManagerImpl {
...
+                .as_mut_node()
+                .set_attribute(AttributeId::Value, value.clone());
+            if let Err(status_code) = res {
+                write.set_status(status_code);
+            }
+            write.set_status(StatusCode::Good);
+            return;
+        }
+
+        write.set_status(StatusCode::BadNothingToDo);
```
oroulet, line 389: @einarmo when adding write to example client, I found out I would recevied BadNotWrittable . Looking at the code, it looks like only callbacks where supported for that SimpleNodeManager... So I added that as an exercise
I am doing something completely wrong or that code was really missing? Don't we have tests for that node manager?

einarmo: I'm a little confused with how you changed the code here. There may have been a bug, but I don't think the old code was so wrong you needed to completely rewrite it?

einarmo: I think your issue may be that there's an unwritten assumption here. If you define a _write_ callback you should also define a _read_ callback. It's assumed that in this case the value is stored externally.

einarmo: Ah no, reading your code, you added writing to the variable even without a callback. I don't want to support this, because IMO it never makes sense in a real server, only in a toy. In any real server you want all writes to somehow have a permanent effect, which means that you need a callback.

oroulet: yes I added writing value to address space without callback, so in memory and no persistence.
 >>  IMO it never makes sense in a real server, only in a toy.
 
 I have been working with ua in at least 15 years now and we do that all the time. There are so many cases where you do not need persistence.
 In that particular case I am writting a PLC emulator and I really do not care about persistence. I want it to restart as per the config

einarmo: Hm.. Alright.., but I'll ask you to change the first bit of the code back, the line where I checked if it was a `Variable` node. Your code is semantically mostly the same, but I don't see a reason for the change.

I still think my argument holds. OPC-UA is an interface, if you are interfacing with nothing then there's no real point to allowing writes at all. A PLC simulator would still want to actually simulate the PLC, meaning it would need to handle writes. Deferring that to the interface layer is, IMO, a misuse of the library.

Essentially you're abusing the fact that OPC-UA kinda looks like your PLC to use it to store simulation values, while what OPC-UA is actually used for is to allow access to values in some _real_ system, simulation or not.

But it's fine, the simple node manager is not really the right choice for most real OPC-UA servers anyway.

oroulet: OK I checked again and you are right we are in fact not doing that for anything else than CI. 
Maybe I can just remove that code and use some CB. In an example we should show best practices. but maybe in that case we should make that method return some clear error or at least log something for the developers.

> our code is semantically mostly the same, but I don't see a reason for the change.

that was to make borrow checker happy, since I reused some variables. but it will become obsolete if I remove that anyway

einarmo: I think you can make that work, you shouldn't be calling `set_attribute` anyway, but the more advanced `set_value_range`, I think...

I gave this some more thought and I think I'm fine with your change. The conflict I think comes from the fact that I consider an OPC-UA server something you build on top of an existing system, to provide an interface. If what you do is instead create something that acts as a client and reacts on changes instead of being directly triggered by writes, it would indeed make sense to store the value.

The users do still have control over whether values should be written, since they have the `AccessLevel` attribute.

I'm at least open for looking into it more. Perhaps we could disallow writes if a read callback was defined? I'm open for opinions on this.

einarmo:
I made some final tidying up, and moved some of the write logic into a dedicated utility method. Small changes, mostly, but in a recent change I added a new way to define type loaders that is a lot less verbose.

einarmo (dismissed):
I'm fine with this now if you're OK with my changes @oroulet.

oroulet:
> I'm fine with this now if you're OK with my changes @oroulet.

Fine too but someone needs to approve. I cannot and maybe you cannot so @svanharmelen maybe?

svanharmelen (approved):

PR #16, opened 2025-01-01
add get_namespace_index to Session

oroulet:


opcua-client/src/session/mod.rs
```diff
@@ -430,4 +431,59 @@ impl Session {
...
+            ))
+        }
+    }
+
+    ///
+    /// read namespace array from server and look for supplied namespace url
+    pub async fn get_namespace_idx(&self, url: &str) -> Result<u16, Error> {
+        let array = self.get_namespace_array().await?;
+        let idx = array.iter().position(|s| s == url).ok_or_else(|| {
+            Error::new(
```
oroulet, line 479: returning an opcua status code error here. that fits well for read_namespace_arry, but maybe not great here. What do you want @einarmo ? use thiserror?

einarmo:
Okay, this works, but the session actually contains a place where we store a `NamespaceMap`, and ways to set it, using [set_namespace_map](https://github.com/FreeOpcUa/rust-opcua/blob/762c821989a628599691fbb490f367f79e870941/opcua-client/src/session/mod.rs#L401).

What if we did:

 - Let `read_namespace_array` store the read namespace map, and return a clone of the `NamespaceMap` (it already contains methods for working with namespaces that are easier to use than just `Vec<String>`, and can be extended).
 - Have `get_namespace_index` just always read from the cached namespace map, and return `None` if no namespace map is present.

That does mean the user must make a dedicated call to fetch the namespace array, but it resolves the question of what the return type from `get_namespace_index` should be (It'd just be an option).

oroulet:
I made get_namespace index update namespace in the background... not sure it is a good idea. sounds easier to me. but as you want. anyway who know what they do anyway would query directly the namespace map

opcua-client/src/session/mod.rs
```diff
@@ -430,4 +431,51 @@ impl Session {
...
+                ),
+            ))
+        }
+    }
+
+    ///
+    /// return index of supplied namespace url
+    /// by reading namespace array from server and looking for supplied namespace url
+    pub async fn get_namespace_index(&mut self, url: &str) -> Result<u16, Error> {
+        let map = self.read_namespace_array().await?;
```
oroulet, line 469: ok I sould get it from the map first and only do query if missing

opcua-types/src/namespaces.rs
```diff
@@ -26,6 +26,23 @@ impl NamespaceMap {
         }
     }
 
+    /// Create a new namespace map from a vec of variant as we get when reading
+    /// the namespace array from the server
+    pub fn new_from_variant_array(array: &[Variant]) -> Self {
```
einarmo, line 31: This should probably return an error if any variant isn't a string, that _is_ an error in this case.

einarmo (commented):
This works, just some nitpicks.

We can add methods for getting directly from the cache without retrieving it from the server later, or make it visible in some way.

opcua-types/src/namespaces.rs
oroulet: yes. I think I originally copied that code from somewhere. Can fix now

oroulet: but once again I need to know what kind of error we should return.
If the project does not define errors, we need to make our own or decide to return the error with statuscode that is used several places

einarmo: In this case you can actually define a unit struct for it, or something like `struct UnexpectedVariantType(usize)` containing the index of the variant that failed or something like that. That's what we do some other places.

oroulet: But simple enums are not compatible with stdeeror and create a lot t of issues isn't it?
Then I would rather use thiserror crate which is anyway a dependency

oroulet: I made an MR: As far as I know this is the correct way to do it, but I do not have so much experience with rust

einarmo (dismissed):
Looks good

oroulet:
@einarmo I pushed right after your approval... so you need to have a look and maybe approve again

einarmo (approved):
Still fine, provided everything passes :+1:

PR #17, opened 2025-01-01
make Error +Send + sync

oroulet:
Suddently I needed this. As far as I understand it is common. It just declare something it already was supporting


einarmo (approved):
Yep, this is probably fair.

PR #18, opened 2025-01-02
add custom errors compatible with std

oroulet:
to be used for internal code and not abuse
 StatusCode errors from protocol

### DIFF
--- a/opcua-core/Cargo.toml
+++ b/opcua-core/Cargo.toml
@@ -28,6 +28,7 @@ url = { workspace = true }
 
 opcua-types = { path = "../opcua-types" }
 opcua-crypto = { path = "../opcua-crypto" }
+thiserror = { workspace = true }
 
 
 [lints.rust]
--- a/opcua-core/src/errors.rs
+++ b/opcua-core/src/errors.rs
@@ -1,0 +1,17 @@
+// OPCUA for Rust
+// SPDX-License-Identifier: MPL-2.0
+
+//!  Rust OpcUa specific errors
+
+use opcua_types::VariantScalarTypeId;
+use thiserror::Error;
+
+/// Rust OpcUa specific errors
+#[allow(missing_docs)]
+#[derive(Error, Debug)]
+pub enum OpcUAError {
+    #[error("Received an expected variant type")]
+    UnExpectedVariantType(Option<VariantScalarTypeId>),
+    #[error("The requested namespace does not exists")]
+    NamespaceDoesNotExist(String),
+}
--- a/opcua-core/src/errors.rs
+++ b/opcua-core/src/errors.rs
@@ -10,7 +10,7 @@ use thiserror::Error;
 #[allow(missing_docs)]
 #[derive(Error, Debug)]
 pub enum OpcUAError {
-    #[error("Received an expected variant type")]
+    #[error("Received an unexpected variant type")]
     UnExpectedVariantType(Option<VariantScalarTypeId>),
     #[error("The requested namespace does not exists")]
     NamespaceDoesNotExist(String),
--- a/opcua-core/src/errors.rs
+++ b/opcua-core/src/errors.rs
@@ -11,7 +11,7 @@ use thiserror::Error;
 #[derive(Error, Debug)]
 pub enum OpcUAError {
     #[error("Received an unexpected variant type")]
-    UnExpectedVariantType(Option<VariantScalarTypeId>),
+    UnexpectedVariantType(Option<VariantScalarTypeId>),
     #[error("The requested namespace does not exists")]
     NamespaceDoesNotExist(String),
 }
--- a/opcua-core/src/lib.rs
+++ b/opcua-core/src/lib.rs
@@ -64,6 +64,7 @@ pub mod constants {
 
 pub mod comms;
 pub mod config;
+pub mod errors;
 pub mod handle;
 
 pub mod messages;
--- a/samples/client.conf
+++ b/samples/client.conf
@@ -9,13 +9,6 @@ verify_server_certs: true
 pki_dir: ./pki
 preferred_locales: []
 default_endpoint: sample_none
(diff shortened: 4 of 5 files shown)


opcua-core/src/errors.rs
```diff
@@ -0,0 +1,12 @@
+use opcua_types::Variant;
+use thiserror::Error;
+
+/// Rust OpcUa specific errors
+#[allow(missing_docs)]
+#[derive(Error, Debug)]
+pub enum OpcUAError {
+    #[error("Received an expected variant type")]
+    UnExpectedVariantType(Variant),
```
einarmo, line 9:
```suggestion
    #[error("Received an unexpected variant type")]
    UnexpectedVariantType(Variant),
```

einarmo, line 9: This should probably contain a `VariantScalarTypeId`, if I understand its purpose correctly. You can easily get that from a variant that isn't `Empty`.

einarmo: Maybe `Option<VariantScalarTypeId>` I guess, to contain the `Empty` case.

oroulet: done

opcua-core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ url = { workspace = true }
 
 opcua-types = { path = "../opcua-types" }
 opcua-crypto = { path = "../opcua-crypto" }
+thiserror.workspace = true
```
svanharmelen, line 31: Didn't see this format being used before (didn't even know it was possible to be honest 😉) but maybe it's best to keep things consistent and write this as:
```suggestion
thiserror = { workspace = true )
```

opcua-core/src/lib.rs
```diff
@@ -66,6 +66,9 @@ pub mod comms;
 pub mod config;
 pub mod handle;
 
+/// Rust OpcUa specific errors
+pub mod errors;
+
```
svanharmelen, line 71:
```suggestion
pub mod config;
pub mod errors;
pub mod handle;

```

opcua-core/src/errors.rs
```diff
@@ -0,0 +1,12 @@
+use opcua_types::{Variant, VariantScalarTypeId};
+use thiserror::Error;
+
+/// Rust OpcUa specific errors
+#[allow(missing_docs)]
+#[derive(Error, Debug)]
+pub enum OpcUAError {
+    #[error("Received an expected variant type")]
+    UnExpectedVariantType(Option<VariantScalarTypeId),
```
svanharmelen, line 9:
```suggestion
    UnExpectedVariantType(Option<VariantScalarTypeId>),
```

opcua-core/src/errors.rs
```diff
@@ -0,0 +1,12 @@
+use opcua_types::{Variant, VariantScalarTypeId};
```
svanharmelen, line 1:
```suggestion
use opcua_types::VariantScalarTypeId;
```

svanharmelen (changes requested):
Some small suggestion to tidy up things, but also some suggested fixes as the code currently doesn't compile.

opcua-core/src/errors.rs
oroulet: how did I manage to commit that....

oroulet:
> Some small suggestions to tidy up things, but also some suggested fixes as the code currently doesn't compile.

thanks. My last commit was obviosuly done far too fast.



But my main question/proposition with that MR is to create that enum for std compatible errors. I was suprised they did not exist in current code base. I thought maybe I was missing something..

samples/client.conf
```diff
@@ -65,10 +66,9 @@ publish_timeout:
 min_publish_interval:
   secs: 0
   nanos: 100000000
-session_timeout: 60000
 performance:
   ignore_clock_skew: false
   recreate_monitored_items_chunk: 1000
-session_name: Rust OPC UA Client
-channel_lifetime: 60000
 recreate_subscriptions: true
```
oroulet, line 72: not sure why that file was changed. I think it it rustfmt playing around. seems to be only ordering so I jsut leave it

svanharmelen: This file is generated when running the tests, so I assume some things have been changed/reordered in other PRs but this file was not added to those PRs. So this is fine and only indicated you've run the test 😉

opcua-core/src/errors.rs
```diff
@@ -0,0 +1,17 @@
...
+//!  Rust OpcUa specific errors
+
+use opcua_types::VariantScalarTypeId;
+use thiserror::Error;
+
+/// Rust OpcUa specific errors
+#[allow(missing_docs)]
+#[derive(Error, Debug)]
+pub enum OpcUAError {
+    #[error("Received an expected variant type")]
```
svanharmelen, line 13:
```suggestion
    #[error("Received an unexpected variant type")]
```

opcua-core/src/errors.rs
```diff
@@ -0,0 +1,17 @@
...
+    UnExpectedVariantType(Option<VariantScalarTypeId>),
```
svanharmelen, line 14: `Unexpected` is one word, so the PascalCase version of this variant should be:
```suggestion
    UnexpectedVariantType(Option<VariantScalarTypeId>),
```

svanharmelen (changes requested):
Just two more little typos and then this should be good to go 👍🏻

oroulet:
@svanharmelen for some reason, you still have some "change request" preventing the merge But I cannot see for what...

svanharmelen:
GitHub doesn't automatically dismiss a review when changes are requested (even if those changes are implemented). So I once your done (which I can see your are) I can take another look and approve the changes if all is good now.

svanharmelen (approved):
LGTM 👍🏻

PR #23, opened 2025-01-07
Workaround for browser future not being Send

einarmo:
This is incredibly hard to diagnose, because the compiler error is useless.

I hate having to box a future in such a critical location, but I can't find a way around it. If someone wants to spend a few hours banging their head against rustc issue 100013, be my guest.

The bug in question is https://github.com/rust-lang/rust/issues/100013. I have a vague intuition that it may be able to dodge this issue by making `UARequest` a generic trait instead of using a GAT. It may be worth trying, as it won't necessarily break usecases.

### DIFF
--- a/lib/tests/integration/browse.rs
+++ b/lib/tests/integration/browse.rs
@@ -617,20 +617,28 @@ async fn test_recursive_browser() {
         .is_some());
 }
 
+// Test that the future returned by `browser.run...` is `Send`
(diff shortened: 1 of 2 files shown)


oroulet:
great! that works!
that stuff is really the dark side of rust...

oroulet (approved):
makes code usable and is not a big change so we must approve

PR #25, opened 2025-01-08
remove useless mut in set_namespaces

oroulet:


oroulet:
@einarmo @svanharmelen do you have any ideas why clippy did not warn about that useless mut?
I wanted to make that method not being mut, was ready to add locks everywhere and then found out it was simply useless...

einarmo (approved):
Clippy doesn't warn about unused `&mut` in arguments like this because in addition to having the meaning "mutable access", `&mut` also means _unique_ access, and sometimes there may be invariants you want to enforce with unique access to a value.

There may be an optional lint for this, but the default clippy lints are supposed to be the ones that are _always_ applicable.

PR #28, opened 2025-01-11
Rename library to async-opcua

einarmo:
Renames only the folders and the libraries themselves, does not change the import path. I also renamed `lib` to `async-opcua`, to be more consistent.


svanharmelen (approved):
Nice 👍🏻

PR #30, opened 2025-01-22
Prepare for initial re-release

einarmo:
Create a couple of scripts using cargo-release to simplify bumping versions and release management. Since we have so many crates it's a little hairy, so I added some scripts to make it more manageable.

I also wrote a rough changelog on the keep-a-changelog format. The old changelog is a little confusing in context, and uses a different format, so I moved it into docs as old_changelog.md.

Set versions on all the crates we intend to publish for now. In the future we may want to find a good way to publish opcua-codegen too, but since it's a binary, not a crate, we will have to look into publishing it as a cargo tool or something along those lines.

The first release will need to be done manually. Not only are we exceeding the per minute limit of new crates (5), but cargo-release doesn't play that well with new releases of interdependent crates. I hope to create a proper release pipeline to automate this in the future.


oroulet:
@einarmo not the right place to discuss it , but you shold rename the repository to async-opcua too so it is possible to google it knowning the crate name

oroulet:
I also hope you have a script so you do not need to do all of that for every release...

oroulet (approved):

einarmo:
The changelog needs a script, but I committed one for bumping all the version numbers, using `cargo-release`.

We should probably rename the crate, yes. `async-opcua-rust` maybe? Calling it just `async-opcua` is maybe a little confusing?

tools/bump-version.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+cargo release version $(./release-package-args.sh) --execute $1
```
svanharmelen, line 2: Would be nice to end the file with a newline...

tools/release-package-args.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+echo "--package async-opcua --package async-opcua-client --package async-opcua-core --package async-opcua-core-namespace --package async-opcua-crypto --package async-opcua-macros --package async-opcua-nodes --package async-opcua-server --package async-opcua-types --package async-opcua-xml"
```
svanharmelen, line 2: Would be nice to end the file with a newline...

svanharmelen:
> Calling it just async-opcua is maybe a little confusing?

Maybe... But then again the Python repo is also just called `opcua-asyncio` and the C one is called `freeopcua` so 🤷🏻‍♂️

einarmo:
> Maybe... But then again the Python repo is also just called `opcua-asyncio` and the C one is called `freeopcua` so 🤷🏻‍♂️

Fair enough. Let's rename the repo.

svanharmelen (approved):

einarmo:
Looks like the default rust versions on github actions went up to 1.84 after I made the PR, fixed it in #32.

PR #35, opened 2025-01-24
Final preparations for release

einarmo:
I got halfway into releasing and realized we should probably have separate READMEs.

This does that, updates documentation in a few places, and removes some default features (it doesn't really make sense to pull in both the server and client by default...). Apparently the json feature on the server was also mostly broken, and halfway removed, so I fixed that too.

Once this is merged I will just publish the remaining crates.

A proper release pipeline is next, but the first release must be done manually.


svanharmelen (approved):
Very nice 👍🏻

PR #37, opened 2025-02-02
Small things

oroulet:


async-opcua-types/src/impls.rs
```diff
@@ -414,3 +414,31 @@ impl Default for IdentityCriteriaType {
         Self::Anonymous
     }
 }
+
+impl WriteValue {
+    /// default constructor with all struct members
```
einarmo, line 419:
```suggestion
    /// Default constructor with all struct members.
```

async-opcua-types/src/impls.rs
```diff
@@ -414,3 +414,31 @@ impl Default for IdentityCriteriaType {
...
+        Self {
+            node_id,
+            attribute_id: attribute_id as u32,
+            index_range,
+            value,
+        }
+    }
+
+    /// return a WriteValue with AttributeId::Value and no index rane,
+    ///  which is the most common case
```
einarmo, line 435:
```suggestion
    /// Create a simple `WriteValue` for writing to the `Value` attribute.
```

async-opcua-types/src/relative_path.rs
```diff
@@ -340,21 +382,6 @@ impl RelativePathElement {
     }
 }
 
-#[derive(Debug, Clone)]
-/// Error returned from parsing a relative path.
-pub enum RelativePathError {
```
einarmo, line 345: Could you, instead of removing this error, just imlement `From<RelativePathError> for OpcUaError`? I don't see a good reason to flatten the error now.

Do feel free to derive `Error` on `RelativePathError`, just using the docstrings as error messages.

einarmo (changes requested):
Mostly good, some nit, and I would rather we not remove the RelativePathError without good cause. Sure, reducing the number of errors is nice, but `OpcUaError` might quickly grow very large if we include all possible error modes from all possible functions in it. It's probably better for the future to have individual error enums where it makes sense, then just include those in `OpcUaError`. Doing so is easy with thiserror anyway.

async-opcua-types/src/relative_path.rs
oroulet: yes might be. I was unsure what the correct approach is here. I just proposed something.
And yes maybe `From<RelativePathError> for OpcUaError` is a good solution. 
I will change that

async-opcua-types/src/errors.rs
```diff
@@ -5,17 +5,41 @@
...
-pub enum OpcUAError {
+pub enum OpcUaError {
     #[error("Received an unexpected variant type")]
     UnexpectedVariantType {
         variant_id: Option<VariantScalarTypeId>,
         message: String,
     },
     #[error("The requested namespace does not exists")]
     NamespaceDoesNotExist(String),
+    #[error("Request returned a StatusCode Error.")]
```
einarmo, line 21:
```suggestion
    #[error("Request returned a StatusCode Error: {0}")]
```

async-opcua-types/src/errors.rs
```diff
@@ -5,17 +5,41 @@
...
+    StatusCodeError(StatusCode),
+    #[error("Generic Error.")]
```
einarmo, line 23:
```suggestion
    #[error("Generic Error: {0}")]
```

async-opcua-types/src/errors.rs
```diff
@@ -5,17 +5,41 @@
...
+    Error(crate::Error),
+    #[error("Function returned a RelativePathError.")]
```
einarmo, line 25:
```suggestion
    #[error("Function returned a RelativePathError: {0}")]
```

einarmo (commented):
Would be nice to include the actual content of the error in the messages, maybe?

einarmo (approved):
LGTM!

PR #39, opened 2025-02-03
Fix bug where certain server capability parameters were cast to u32 instead of u16

gudjont89:
According to the OPC-UA standard (https://reference.opcfoundation.org/Core/Part5/v104/docs/6.3.2) the following parameters should have the type uint16:
- MaxBrowseContinuationPoints
- MaxHistoryContinuationPoints
- MaxQueryContinuationPoints
In the current implementation of the library, these parameters are cast to u32. This causes a type mismatch bug when connecting from an OPC-UA client on a B&R PLC to a server using the library. The fix was verified on that B&R PLC, but has not been tested on other PLCs.

### DIFF
--- a/async-opcua-server/src/node_manager/memory/core.rs
+++ b/async-opcua-server/src/node_manager/memory/core.rs
@@ -274,16 +274,16 @@ impl CoreNodeManagerImpl {
                 (limits.max_array_length as u32).into()
             }
             VariableId::Server_ServerCapabilities_MaxBrowseContinuationPoints => {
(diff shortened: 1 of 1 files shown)


einarmo (approved):
Huh. These values being u16 honestly is bizarre to me. It's absolutely not inconceivable that you would want to support more than 65k continuation points on a sufficiently large server. I guess if you want that you're stuck with setting the value to 0? (We don't currently support this I think...). All that for a theoretical saving of 4 bytes...

Either way, thanks for your contribution!

PR #43, opened 2025-02-20
Refactor FromXml to use same context and errors as binary and json

einarmo:
Required putting the Aliases table and IndexMap into Context, but that may make sense anyway, if we want to parse binary strings in XML schemas (yes this is, in theory, a thing...).

This is the first step to reworking FromXml into XmlDecodable and adding XmlEncodable, to make XML a proper first-class citizen like JSON and binary.

I did this first. The rework into XmlDecodable is going to be complicated enough without this refactor on top.

Next step is to build a generic XML stream deserializer, probably using quick-xml. We'll probably keep roxmltree for async-opcua-xml for the time being, since that library deals with much more complicated structures than what we find in OPC-UA XML.


svanharmelen (approved):
Much cleaner, nice 👍🏻

PR #44, opened 2025-02-23
fix feature enabling bug for json

oroulet:


einarmo (approved):
We need a test in CI for compiling without the json feature. Will look into it.

PR #45, opened 2025-02-24
Verify that we can build without default features

einarmo:
This has caused issues a few times now, we should check for it in CI.

### DIFF
--- a/.github/workflows/ci_clippy.yml
+++ b/.github/workflows/ci_clippy.yml
@@ -9,3 +9,6 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy -- -D warnings
+      
+      - name: No default features
+        run: cargo clippy --no-default-features -- -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_clippy.yml
```diff
@@ -9,3 +9,6 @@ jobs:
 
       - name: Run clippy
         run: cargo clippy -- -D warnings
+      
+      - name: No default features
+        run: cargo clippy --no-default-features -- -D warnings
```
oroulet, line 14: will clippy catch any build error?

einarmo: Yes. `-- -D warnings` is passing deny warnings to `rustc`.

einarmo: Notably, the warning I fix in this PR is produced with `cargo check --no-default-features`, as well as `cargo clippy --no-default-features`.

PR #47, opened 2025-02-26
Xml encoding

einarmo:
This is a major rewrite of XML parsing, adding support for XML serialization, and bringing it up to the same level as JSON and binary. I also fixed a few bugs I hit on in a few places. There are still some things remaining:

 - Unions
 - XML in binary/json encoded extension objects.
 - Proper support for the XmlElement variant (currently it doesn't work properly for XML). Note that this variant type is deprecated.

This PR is already way too big. There was no real way to make it smaller, since once you start replacing FromXml there's no way to stop while still having the library compile.

The PR is split into a few smaller commits, that incrementally add features, at least as much as was possible while still having each commit compile. The line count may be a little misleading, since a fair bit of it is generated, but this is a _big_ PR, just because there are so many changes.

I want to, in the future, use `quick-xml`s serde feature to replace `roxmltree` entirely, but this is currently not possible, as there is no way to capture raw XML, which we need to in order to defer loading of values for later, once `Context` is available.

We may also want to create an attribute-macro to automatically derive all the different encoding macros depending on input type, since just the derives are getting a bit verbose.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,6 +347,7 @@ dependencies = [
  "chrono",
  "hashbrown 0.15.2",
  "log",
+ "percent-encoding-rfc3986",
  "regex",
  "serde_json",
  "struson",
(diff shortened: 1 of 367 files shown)


svanharmelen (approved):
I must admit I mainly looked at the first two commits (thanks for splitting them) and only skimmed through the last one 😉 But looks good to me 👍🏻

einarmo:
Thanks for the review! I hope and believe this will be the last PR of this size, (excluding any more generated code in the future).

PR #49, opened 2025-02-27
Handle null values in XML as well

einarmo:
This is the same as what we do for JSON, we make it possible to leave out values that are null, which is important for certain special values like UAString, where an empty value is different from a null value.

I considered making this a separate trait, but it would be annoying to have to derive a trait that does nothing most of the time, and without specialization there's no better way. We also can't put it on BinaryEncodable since that trait isn't implemented for `Option`, and can't really be at the moment. It's now effectively duplicated between JsonEncodable and XmlEncodable, which is fine.

### DIFF
--- a/async-opcua-types/src/xml/encoding.rs
+++ b/async-opcua-types/src/xml/encoding.rs
@@ -55,6 +55,12 @@ pub trait XmlEncodable: XmlType {
         writer: &mut XmlStreamWriter<&mut dyn Write>,
         context: &Context<'_>,
     ) -> EncodingResult<()>;
+
+    /// This method should return `true` if the value is default
+    /// and should not be serialized.
+    fn is_null_xml(&self) -> bool {
+        false
+    }
 }
 
 /// Extensions for XmlStreamWriter.
--- a/async-opcua-macros/src/encoding/xml.rs
+++ b/async-opcua-macros/src/encoding/xml.rs
@@ -102,12 +102,16 @@ pub fn generate_xml_encode_impl(strct: EncodingStruct) -> syn::Result<TokenStrea
         if field.attr.optional {
             body.extend(quote! {
                 if let Some(item) = &self.#ident {
-                    stream.encode_child(#name, item, ctx)?;
+                    if !item.is_null_xml() {
+                        stream.encode_child(#name, item, ctx)?;
+                    }
                 }
             });
         } else {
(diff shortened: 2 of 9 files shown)


async-opcua-types/src/xml/encoding.rs
```diff
@@ -55,6 +55,12 @@ pub trait XmlEncodable: XmlType {
         writer: &mut XmlStreamWriter<&mut dyn Write>,
         context: &Context<'_>,
     ) -> EncodingResult<()>;
+
+    /// This method should return `true` if the value is default
+    /// and should not be serialized.
+    fn is_null_xml(&self) -> bool {
+        false
```
svanharmelen, line 62: Guess this method will never return `true`. Is that expected?

einarmo: This is the trait default, it's overridden in certain types.

svanharmelen: As it's defined in a trait, wouldn't it be better to not implement it as a default `false` but require it to be implemented by anyone that implements this trait for a type?

einarmo: We _could_ do that, I guess. It would be a lot of noise though, and most implementations would simply be `false`.

If we do this, I'd rather do it separately, since we would want to do the same for JSON. Maybe combine it with moving this to a separate trait?

svanharmelen: Just wondered if this could cause unexpected behavior were a type forgets to implement/overwrite the default. But as the default is `false` I guess that would be unlikely.

svanharmelen (approved):

einarmo:
It's not an entirely unreasonable thought, and it's a danger of default implementations. In this case, leaving it `false` is almost never wrong, however, since the vast majority of types are fine with being deserialized to null.

The only real exception is `UAString`, which has the unfortunate property that

```xml
<String><String />
```

and 

```xml
<String />
```

mean different things. One is an empty string, the other is a null string. Since the type only governs the _content_ of the tag in our XML serializer, it means we can deserialize the latter, and serialize that back as the first without this feature.

For _every other type_, as well as for all types in JSON, the only advantage is strict standards compliance (the standard wants you to leave out "null" values in some cases), and data size reduction.

PR #50, opened 2025-02-28
Create a unified macro to derive all the encodable traits

einarmo:
The derives are getting very verbose, which is especially bad for users that want to create their own custom structs. This adds an attribute macro that simply derives these traits as appropriate.

I also found an issue with the XML parsing and custom codegen that required an unpleasant workaround.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Mark generated files as generated, for github
+Cargo.lock linguist-generated=true
+async-opcua-types/src/generated/** linguist-generated=true
+async-opcua-core-namespace/src/generated/** linguist-generated=true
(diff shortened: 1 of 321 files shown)


.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Mark generated files as generated, for github
+Cargo.lock linguist-generated=true
+async-opcua-types/src/generated/** linguist-generated=true
+async-opcua-core-namespace/src/generated/** linguist-generated=true
+async-opcua-core-namespace/src/events/generated.rs linguist-generated=true
```
svanharmelen, line 5: Newline is missing...

PR #52, opened 2025-03-03
Support encoding and decoding XML unions

einarmo:
Not a complicated change. I also remove all traces of FromXml in this PR, since those were left behind after the refactor. We no longer use them, and don't want the burden of maintaining the macros.

The encoding/decoding impl is very similar to the one for JSON. Again ignoring the switch field on decoding, since the information is present in the tag name.


oroulet (approved):
did not review closely but that seems quite a bunch of cleaning up and does not breaks tests so...

PR #53, opened 2025-03-04
Improve handling of unions in CustomStruct

einarmo:
We didn't properly handle null unions, we do so now.

Also add a bunch of tests for all the different encodings, including union round-trips now that we have native union support for all the different encodings.

I also added a type to make it more convenient to implement `TypeLoader`, since the support code for the test got a little annoying.


svanharmelen (approved):
👍🏻

PR #55, opened 2025-03-05
Support for XmlElement in XML bodies.

einarmo:
This is technically deprecated in the OPC-UA standard, but it's still good to have support for it.

The old solution just used a type-alias for UAString, which is fine for both JSON and binary, but won't work properly for XML. The lack of a concrete, unique type also caused issues with variant.

This adds a newtype for it instead, just a wrapper around UAString.

Reading the contents of an element as raw XML was quite hairy, but the solution I ended up with seems to work fine.


svanharmelen (approved):
A bit hairy indeed, but nice to have support for this one as well.

PR #58, opened 2025-03-07
Tidy up the Cargo.toml file (ordering and comments)

svanharmelen:
Not necessarily needed, but feels like a nice cleanup...


svanharmelen:
Yeah its indeed a matter of keeping an eye on this when reviewing. Not linting tools available as far as I know...

PR #60, opened 2025-03-11
Add a schema cache for code generation

einarmo:
Add a centralized schema cache when doing code generation, to avoid loading the same schema twice.

This is not really a big improvement yet, but it's increasingly necessary. It also now helps by automatically identifying dependent nodesets, it also makes it easier to name the schemas, and may help us simplify future code gen config.

Note that as intended, nothing has changed in the generated code.

### DIFF
--- a/async-opcua-codegen/src/input/nodeset.rs
+++ b/async-opcua-codegen/src/input/nodeset.rs
@@ -1,0 +1,171 @@
+use std::collections::{HashMap, HashSet};
+
+use opcua_xml::{
+    load_nodeset2_file,
+    schema::{opc_ua_types::Variant, ua_node_set::UANodeSet},
+    XmlElement,
+};
+
+use crate::CodeGenError;
+
+use super::SchemaCache;
+
+pub struct NodeSetInput {
+    pub xml: UANodeSet,
+    pub aliases: HashMap<String, String>,
+    pub uri: String,
+    pub required_model_uris: Vec<String>,
+    /// Map from numeric ID to documentation link.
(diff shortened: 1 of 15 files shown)


svanharmelen:
image

You can (and probably should) prevent that by doing a git pull or rebase on your local branch before pushing (or after pushing, but then you need a force push to update it here). Merging master into your feature branch like this is kind of nasty (even thought it will be "hidden" when squash merging) 😊

async-opcua-codegen/src/input/nodeset.rs
```diff
@@ -0,0 +1,171 @@
...
+                .iter()
+                .map(|a| (a.alias.clone(), a.id.0.clone()))
+                .collect::<HashMap<_, _>>()
+        });
+        let Some(models) = nodeset.models.as_ref() else {
+            return Err(CodeGenError::missing_required_value("Models"));
+        };
+
+        if models.models.len() > 1 {
+            println!("Warning, multiple models found in nodeset file, this is not supported, and only the first will be used.");
```
svanharmelen, line 95: Do you intended these to be `println`, not `log::debug` (or maybe `log::info`)? Noticed this was in the code already, but it just caught my eye...

einarmo:
> You can (and probably should) prevent that by doing a git pull or rebase on your local branch before pushing (or after pushing, but then you need a force push to update it here). Merging master into your feature branch like this is kind of nasty (even thought it will be "hidden" when squash merging) 😊

Yeah, I usually do, I just merged before I thought about it... I can rebase this one.

svanharmelen (approved):

PR #61, opened 2025-03-12
Use NodeSet2 XML for codegen

einarmo:
Before you panic at the line count, it's mostly just the two XML schemas, and the rest is mostly generated code, but this _is_ a pretty complex change. The current commits are intended to be separate and should divide this up a little, but in practice changing codegen tends to break everything until it's completely fixed.

Technically, BSD files are deprecated, and some standards don't define them. What we're supposed to do instead is use NodeSet2 xml files. This has some advantages, but also the major disadvantage that it is _really_ complicated.

An interesting side-effect of this is that the BSD schema refers only to _actual_ types, i.e. how the types are represented in binary. The XML file refers to their real types, which in some cases are more high level. For most of these I simply add a type alias (which may or may not actually be sufficient for XML...), but for NumericRange specifically I transition to using that type directly in the schema, which is quite neat.

We also lose a few abstract types, which seems to fix a few bugs, actually.

The generated types should _mostly_ be the same, except for the change I mentioned above about using more accurate type aliases. We may want to replace these with newtypes or even custom types in the future.

I will note that doing this is way harder than using BSD files, since we now have to potentially step through multiple XML schemas to find the information we need. For example if a type in schema A is a subtype of a type in schema B which is a subtype of a type in the core namespace, we need to traverse through B to figure out what kind of type `A` really is, and what fields it has.

I ended up pulling the latest schemas from github, so they are updated. In the future we may want to reference them via URL instead, but that felt like an unrelated change for this PR.

There is a potential concern I realized with this. `Vec<Duration>` is encoded as list of `f64` in JSON and binary, which is correct, but I think it is encoded as

```
<Double>...</Double>
<Double>...</Double>
```

in XML, while I think it should be

```
<Duration>...</Duration>
<Duration>...</Duration>
```

we may need to build newtypes for all of these, which may be a good idea to begin with, but we probably want support in the derive macros for this.

Could be kind of cool, actually, to have `Duration` be `UADuration(std::time::Duration)`.


oroulet (approved):

oroulet:
I cannot say I reviewed in detail but code quality looks good and you do not break things so...

PR #62, opened 2025-03-14
Add `impl_encoded_as` macro

einarmo:
This is relatively common, where we want to encode a type by first converting it to another type. This adds a macro to implement encoding for these types. Makes code more compact. I found it could be used nicely in generated option sets from the core namespace.


svanharmelen (approved):
👍🏻

PR #67, opened 2025-03-27
Handle zero-padding in legacy encrypted token secrets

mparadzik:
First of all thanks for the work you put into the library, it's been a pleasure working with it.

While investigating interop issues between async-opcua server and legacy application client that uses open62541 when using `SignAndEncrypt` security mode, I discovered that open62541 pads token secret with zero bytes before encrypting, while async-opcua does not account for this when decrypting.

[Relevant section ](https://github.com/open62541/open62541/blob/cae846b1e4504af6ee4953a131bb5fe96eaf38af/src/client/ua_client_connect.c#L283C1-L292) in open62541 code (where I took comments from).

This PR adds a check for such padding and trims it before proceeding with password extraction, which makes server compatible with clients that do padding before secret encryption.

### DIFF
--- a/async-opcua-crypto/src/user_identity.rs
+++ b/async-opcua-crypto/src/user_identity.rs
@@ -167,16 +167,39 @@ pub fn legacy_password_decrypt(
         // Decrypt the message
         let src = secret.value.as_ref().unwrap();
         let mut dst = vec![0u8; src.len()];
-        let actual_size = server_key
+        let mut actual_size = server_key
             .private_decrypt(src, &mut dst, padding)
             .map_err(Error::decoding)?;
 
         let mut dst = Cursor::new(dst);
         let plaintext_size = read_u32(&mut dst)? as usize;
+
+        /* Remove padding
+         *
+         * 7.36.2.2 Legacy Encrypted Token Secret Format: A Client should not add any
+         * padding after the secret. If a Client adds padding then all bytes shall
+         * be zero. A Server shall check for padding added by Clients and ensure
+         * that all padding bytes are zeros.
+         *
+         */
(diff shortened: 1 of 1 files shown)


async-opcua-crypto/src/user_identity.rs
```diff
@@ -167,16 +167,39 @@ pub fn legacy_password_decrypt(
...
+        if actual_size > plaintext_size + 4 {
+            let padding_bytes = &dst[plaintext_size + 4..];
+            /*
+             * If the Encrypted Token Secret contains padding, the padding must be
+             * zeroes according to the 1.04.1 specification errata, chapter 3.
+             */
+            if !padding_bytes.iter().all(|&x| x == 0) {
+                return Err(Error::decoding("Non-zero padding bytes in decrypted password"));
+            } else {
+                dst = dst[..plaintext_size + 4].to_vec();
```
einarmo, line 195:
```suggestion
                dst.truncate(plaintext_size + 4);
```

just to avoid an unnecessary reallocation.

einarmo (commented):
Such an OPC-UA standard thing "should not add padding".

Thanks! I just have one tiny suggestion to avoid reallocating the decrypted secret.

einarmo:
Looks like you also need to run `cargo fmt --all`.

mparadzik:
Applied suggestion to truncate the vector, and also fixed formatting (my bad).

async-opcua-crypto/src/user_identity.rs
mparadzik: Done, thanks for suggestion

einarmo (approved):
Thanks!

PR #68, opened 2025-04-01
Use log for logging in codegen

einarmo:
Also set the log level to debug by default. Users can override it if they want, but for must usage we just want all the logs always.

There isn't a ton of logging yet, but we should standardize on using `log` for everything.

### DIFF
--- a/async-opcua-codegen/src/input/nodeset.rs
+++ b/async-opcua-codegen/src/input/nodeset.rs
@@ -3,7 +3,7 @@ use std::{
     sync::OnceLock,
 };
 
-use log::warn;
+use log::{info, warn};
 use opcua_xml::{
     load_nodeset2_file,
     schema::{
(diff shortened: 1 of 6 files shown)


oroulet:
talking about og. I jsut found out that repository is using some custome env varaible RUST_OPCUA_LOG. I did not look into the details but that seems quite wrong. Why should a library override standard rust program behaviour?
It just make things very confusing and I spend at least one hours trying to figure out why RUST_LOG was not repected for this crate while all the others worked fine...

oroulet (approved):

oroulet:
what is the poit of the init_console_logging think?
I admit I have alway been using tracing and it has worked nicelly

oroulet:
but log is more general I understand

einarmo:
Adding tracing is actually on my TODO somewhere. It's more for measuring the timing of events, and then a possible sink for tracing is to write it to console. There is definitely some overlap. If we add tracing it will be used differently than logging:

 - Log warnings, general information about important events, and errors.
 - Add tracing systematically to each individual service, and potentially in a few other specific places.

Issue #73, opened 2025-04-03
linking fails for fuzz


AiyionPrime:
This might be an archlinux issue, just as heads up.

```console
[...]
  Downloaded rustix v0.38.43
  Downloaded jiff v0.2.4
  Downloaded tokio v1.43.0
  Downloaded ring v0.17.8
  Downloaded 109 crates (11.2 MB) in 1.39s (largest was `ring` at 4.2 MB)
   Compiling async-opcua-fuzz v0.0.0 (/home/aiyion/Documents/async-opcua/fuzz)
error: linking with `cc` failed: exit status: 1
  |
[...]

PR #74, opened 2025-04-03
Make fuzz targets build without the nightly feature

einarmo:
This was a workaround that didn't work that well. I don't love the solution, since it makes the code harder to read, but it works, and now you can run `cargo build` without it failing on master.


AiyionPrime (approved):
Just verified, this indeed closes #73 .
Thanks!

oroulet (approved):

oroulet:
I approved but honestly I know nothing about these things

PR #75, opened 2025-04-03
Make a note on choice of namespace_uri

einarmo:
This caused some confusion. I don't know how much we can really do to prevent issues here, but we can at least make a comment in the sample, since most people will start from here.


AiyionPrime (approved):
Looks good, thanks :)

oroulet (approved):

PR #77, opened 2025-04-04
Add tracing to the server

einarmo:
This adds layered tracing to the server, and cleans up what we log a bit (logging the message chunks is insane).

Tracing is nice and automatic, though you do have to be careful about tracing over await points. This should make it possible (with the correct subscriber) to produce structured spans for each request on the server. When not using tracing, it provides decent debug-level logs with some level of detail.

We trace connection establishment, requests on the top level, as well as per call to a node manager. Node managers can do their own tracing and logging, which will automatically be part of the top level span.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -368,6 +368,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
+ "tracing-futures",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


svanharmelen:
Just a quick FYI... I'm on holiday for 4 weeks with little to no internet. So I'm still here, just not able to review and/or approve 😉

einarmo:
> Just a quick FYI... I'm on holiday for 4 weeks with little to no internet. So I'm still here, just not able to review and/or approve 😉

It's all good. No hurry. This came up and I figured it might be useful.

oroulet (approved):

PR #86, opened 2025-04-25
Add unused pub lint

einarmo:
This lint is handy, as it notifies us of items marked `pub` that are not actually public in the crate itself. This is useful, since a common source of issues is that we forget to expose some component that we want to expose.

I also made a tiny, related refactor of the security policies in the crypto library, essentially just moving the constants for each policy onto a type, as a trait, instead of as a module each. We may be able to use a `dyn SecurityPolicy` or something here in the future, to avoid needing the enum, which would be quite nice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,11 @@ members = [
   "fuzz",
 ]
 
+[workspace.lints.rust]
+missing_docs = "warn"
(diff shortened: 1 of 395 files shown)


svanharmelen (approved):
Very nice one 👍🏻

PR #88, opened 2025-04-25
Improve legacy secrets

einarmo:
Some misc code improvement around legacy secrets. I want to add support for issuedtoken secrets on the server and client, which needs to reuse the machinery that already exists for username authentication.

We also didn't quite follow the standard here, but we should now do so more explicitly.


svanharmelen (approved):
Looks good, but why the name "legacy secret"?

einarmo:
> Looks good, but why the name "legacy secret"?

https://reference.opcfoundation.org/Core/Part4/v105/docs/7.41.2.2, it's roughly how the standard refers to these.

PR #89, opened 2025-04-25
Update client documentation for monitoring the event loop

benidi:
I have updated the client documentation for Monitoring the eventloop to use enter() and Box::pin.


docs/client.md
```diff
@@ -256,11 +256,14 @@ Using `event_loop.spawn` is convenient if you do not care what the session is do
 
 You can watch these events yourself by using `event_loop.enter`, which returns a `Stream` of `SessionPollResult` items.
 
+If you're running inside an async context (like a `#[actix_web::main]` function, async handler, etc) use `tokio::task::spawn_blocking`
```
einarmo, line 259: This is is a _really_ bad practice. I don't want to recommend it.

Issue #102, opened 2025-05-09
Invalid OPCUA sequence numbers sent by the server


evanjs:
Note: this issue also appeared on `locka99/opcua` under https://github.com/locka99/opcua/issues/190

This can be seen when using e.g. the web UI for Ignition from Inductive Automation, as noted in the original issue.

I noticed the behavior again when migrating to this repository, and wanted to raise the issue here before I forget, make sure I'm not missing anything simple.

I did notice we've done some work on sequence numbers lately
https://github.com/FreeOpcUa/async-opcua/commit/83d12f0367e3ec250bb90f604338f5f32d641278

I did not verify the behavior before and after this commit

einarmo:
I haven't seen this behavior. Do you have data on what request it starts with? It does indeed look like a missing package.

evanjs:
Right, sorry, forgot the main detail of when this appears in the logs
Didn't realize the original issue also lacks much beyond "this happens", haha oops

I can observe this from the Ignition client when subscribed via the Ignition OPC Quick Client
I _believe_ Ignition Designer results in the same behavior as well, but the Quick Client is much easier to use/test 

___
## How to Reproduce
Assuming you have connected to an OPCUA Server using the Ignition client:

Navigate to the `OPC Quick Client` page

`Config` (Sidebar) > `OPC CLIENT` (Header/Group) > `OPC Quick Client`
URL path: `/web/config/opc.quickclient`

Interacting with a tag from the address space browser using the associated action column (`[s]`) should add the tag to the `Subscription 1` interface below

Then, `Subscription name` and `Rate (ms)` can be adjusted before the subsciption is saved using `Set` 

___
### How often does this happen?

On this page, if I subscribe to a tag, I see the sequence mismatch after... something like 4 minutes with my current settings
___
### Is it consistent?
It does appear to be consistent once triggered
```
OpcUaSubscriptionManager | 09May2025 14:13:40 | [id=1] expected sequence=6, received sequence=7. Calling Republish service...
OpcUaSubscriptionManager | 09May2025 14:09:32 | [id=1] expected sequence=5, received sequence=6. Calling Republish service...
OpcUaSubscriptionManager | 09May2025 14:05:24 | [id=1] expected sequence=4, received sequence=5. Calling Republish service...
```
___
### Did this happen before?
It was definitely more prominent with `locka99/opcua` (something like every 2 minutes after I worked around the Keepalive stuff misbehaving), but I _do_ still notice the behavior here

[...]

PR #103, opened 2025-05-10
Reuse sequence numbers for keep alive messages

einarmo:
This is related to #102, where a client is producing a warning due to sequence numbers not being the expected value. We don't currently verify these in the client, which we may do in the future.

The issue appears to be that the keep-alive messages should not increment the sequence number, but the standard is _really_ vague on this stuff. The only real source is this haphazard list of rules, https://reference.opcfoundation.org/Core/Part4/v104/docs/5.13.1.1, of which the relevant part is 

> Each keep-alive [Message](https://reference.opcfoundation.org/search/17?t=Message) is a response to a [Publish](https://reference.opcfoundation.org/search/17?t=Publish) request in which the [notificationMessage](https://reference.opcfoundation.org/search/17?t=notificationMessage) parameter does not contain any [Notifications](https://reference.opcfoundation.org/search/17?t=Notifications) and that contains the sequence number of the next [NotificationMessage](https://reference.opcfoundation.org/search/17?t=NotificationMessage) that is to be sent

Also, it seems like after this, the rest of the page treats `NotificationMessage` as a separate concept from `notificationMessage` (not even a little confusing), where the latter is an OPC-UA type, and the former is a `notificationMessage` containing at least one notification.

Except, _status_ messages do not contain notifications, but _do_ increment the sequence number, so the standard is wrong!

Either way, I looked at the reference SDK, and it does it this way, though in a very roundabout manner.

### DIFF
--- a/async-opcua-core/src/handle.rs
+++ b/async-opcua-core/src/handle.rs
@@ -35,6 +35,11 @@ impl Handle {
         next
     }
 
+    /// Peek the next value of the handle, without incrementing.
+    pub fn peek_next(&self) -> u32 {
+        self.next
+    }
+
     /// Set the next handle value manually.
     pub fn set_next(&mut self, next: u32) {
         self.next = next;
--- a/async-opcua-server/src/subscriptions/session_subscriptions.rs
+++ b/async-opcua-server/src/subscriptions/session_subscriptions.rs
@@ -602,6 +602,7 @@ impl SessionSubscriptions {
             // Get notifications and publish request pairs while there are any of either left.
             while !self.publish_request_queue.is_empty() {
                 if let Some(notification_message) = subscription.take_notification() {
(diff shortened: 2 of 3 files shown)


svanharmelen (approved):

evanjs:
Unforunately, I cannot share any screenshots of further warning messages in the logs

... this is because they no longer seem to appear when this fix is applied 🎉 

This is _with_ my initial repro as suggested by @einarmo
Max Keepalive Count: 30,000 -> 3
And a subscription on Ignition's OPC Quick Client to `v5`

Based on this scenario alone, I do not see the issue reoccuring

Thank you!
___
Still a **_tiny_** bit paranoid based on my own diagnostics into this issue before, but I'll keep an eye on it
For now, things seem to be working as expected, with zero (client-side) log spam
___
This should close #102

PR #105, opened 2025-05-12
Change TcpConfig and CertificateValidation to public.

digitizer0:
The TcpConfig and CertificateValidation structs would be good to have as public when creating custom configuration file in eg. toml.


evanjs:
I was just trying to test this https://github.com/FreeOpcUa/async-opcua/issues/102#issuecomment-2867618871 the other day and would have benefited from this change in visibility.

einarmo:
Closed via #106

PR #107, opened 2025-05-19
Some refactoring of the session struct

einarmo:
- Move a few fields exclusively into the AsyncSecureChannel struct.
 - Make it possible to create a secure channel directly from the session builder.
 - Give the secure channel proper ownership of the authentication token. The current approach is a little confusing.
 - Rename `SessionInfo` to `EndpointInfo` which is a lot more correct.

I'm experimenting a little with creating a sample that doesn't use the `Session` struct. I want to see if I can make a _really_ reliable OPC-UA client with persistent state, so that you can run an application that can keep going through server restarts, client restarts, and solar flares.

### DIFF
--- a/async-opcua-client/src/session/client.rs
+++ b/async-opcua-client/src/session/client.rs
@@ -29,8 +29,8 @@ use opcua_types::{
 };
 
 use super::{
-    connection::SessionBuilder, process_service_result, process_unexpected_response, Session,
-    SessionEventLoop, SessionInfo,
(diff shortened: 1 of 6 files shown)


oroulet (approved):

oroulet:
planning to handle reconnection automatically?

svanharmelen (approved):

einarmo:
We already have some level of automatic reconnection. I hope to make it more flexible and powerful in the future, but I'm still exploring what we can do.

PR #109, opened 2025-05-20
Pause publish loop when receiving BadNoSubscription.

einarmo:
Reading the standard a bit more, it seems like
`BadSubscriptionIdInvalid` isn't a valid service-level status code for `Publish`, so we don't need to consider that one here. As for `BadNoSubscription`, if this happens we are _probably_ in the process of creating a new subscription, so we can pause the automatic publishing until it is created. For safety we also unpause if we successfully receive a publish response.

@svanharmelen what do you think? This is the only real way I can see to handle this, in this case. We do have mechanisms to auto-recreate subscriptions on reconnect, but I'd not be too surprised to see this response happen before we actually finish that auto-recreate process.

### DIFF
--- a/async-opcua-client/src/session/services/subscriptions/event_loop.rs
+++ b/async-opcua-client/src/session/services/subscriptions/event_loop.rs
@@ -29,7 +29,10 @@ pub(crate) struct SubscriptionEventLoop {
     last_external_trigger: Instant,
     // This is true if the client has received BadTooManyPublishRequests
     // and is waiting for a response before making further requests.
-    is_waiting_for_response: bool,
+    waiting_for_response: bool,
+    // This is true if the client has received a no_subscriptions response,
+    // and is waiting for a manual trigger or successful response before resuming publishing.
+    no_active_subscription: bool,
 }
 
 impl SubscriptionEventLoop {
@@ -50,7 +53,8 @@ impl SubscriptionEventLoop {
             last_external_trigger,
             trigger_publish_recv,
             session,
-            is_waiting_for_response: false,
+            waiting_for_response: false,
+            no_active_subscription: false,
         }
     }
 
@@ -70,7 +74,7 @@ impl SubscriptionEventLoop {
                     // if there are no active subscriptions. In this case, simply return the
                     // non-terminating future.
                     let next_tick_fut = if let Some(next) = next {
-                        if slf.is_waiting_for_response && !futures.is_empty() {
+                        if slf.waiting_for_response && !futures.is_empty() {
                             Either::Right(futures::future::pending::<()>())
                         } else {
                             Either::Left(tokio::time::sleep_until(next.into()))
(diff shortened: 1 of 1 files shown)


async-opcua-client/src/session/services/subscriptions/event_loop.rs
```diff
@@ -30,6 +30,9 @@ pub(crate) struct SubscriptionEventLoop {
     // This is true if the client has received BadTooManyPublishRequests
     // and is waiting for a response before making further requests.
     is_waiting_for_response: bool,
+    // This is true if the client has received a no_subscriptions response,
+    // and is waiting for a manual trigger or successful response before resuming publishing.
+    is_paused: bool,
```
svanharmelen, line 35: Maybe something like this:

```suggestion
    waiting_for_response: bool,
    // This is true if the client has received a no_subscriptions response and is waiting
    // for a manual trigger or successful response before resuming publishing.
    no_active_subscription: bool,
```

_Note that I updated `is_waiting_for_response` on purpose as well_

async-opcua-client/src/session/services/subscriptions/event_loop.rs
```diff
@@ -51,6 +54,7 @@ impl SubscriptionEventLoop {
             trigger_publish_recv,
             session,
             is_waiting_for_response: false,
+            is_paused: false,
```
svanharmelen, line 57:
```suggestion
            waiting_for_response: false,
            no_active_subscription: false,
```

async-opcua-client/src/session/services/subscriptions/event_loop.rs
```diff
@@ -162,16 +168,13 @@ impl SubscriptionEventLoop {
...
-                                        | StatusCode::BadSubscriptionIdInvalid => {
-                                            // TODO: Maybe do something here? This could happen when
-                                            // subscriptions are in the process of being recreated.
-                                            // Make sure to avoid race conditions.
-                                            session_error!(
+                                        StatusCode::BadNoSubscription => {
+                                            session_debug!(
                                                 slf.session,
-                                                "Publish response indicates subscription is dead",
+                                                "Publish response indicates that there are no subscriptions. Pausing publish loop until a new subscription is created, or we receive a successful publish response.",
```
svanharmelen, line 174: This would be a very long wrapping logline in the logfile, I prefer to keep these a bit more concise:

```suggestion
                                                "Publish response indicates that there are no subscriptions",
```

svanharmelen (approved):
I'm not sure if this adds much value, but I guess it doesn't hurt either. Have some minor suggestions about the chosen wording...

In my own loop logic I used to make some calls when either of these events came in, but last week during debugging an issue I noticed these events indeed only come in when you are making calls yourself (e.g. to transfer a subscription), so I reverted that code back to only log that the event is received to prevent triggering actions to recover from a "lost" subscription which was actually work in progress to create a new subscription.

So I ended up with a never ending loop of subscription transfers and/or recreates 😅

einarmo:
> I'm not sure if this adds much value, but I guess it doesn't hurt either. 

Yeah... I also realized there isn't a ton that makes sense to do here. Trying to recreate the subscription would be working around a server that just deleted the subscription for no clear reason? That sounds like a dangerous thing to build in automatically.

svanharmelen (approved):

PR #111, opened 2025-05-22
remove a println

mrxdst:
I don't think its supposed to be there.


einarmo:
Oops! I think there may be a lint for println, maybe something to look into...

PR #113, opened 2025-06-03
fix: documentation for minimum publishing and sampling limits

Andy3189:
Small documentation fix for `min_sampling_interval_ms` and `min_publishing_interval_ms` in `SubscriptionLimits` from seconds to milliseconds to align with their names and usage inside the library.


einarmo:
Looks like we're on a newer rust version now and need to deal with new clippy lints. I'll make a PR to fix that.

PR #121, opened 2025-07-04
Support for the OfType operator

einarmo:
IIRC I didn't do this originally because it was hard to fit into the server architecture. We're a bit more settled now, and it is possible to implement.

Still, it's a little annoying to do, since now evaluating event filters requires the type tree, meaning that we need to share that somehow. I made it so that SessionSubscriptions holds a reference to a type tree for the user it belongs to, which _should_ be sufficient. It does mean we take a read lock on the type tree whenever we notify about subscriptions. The hope is that most servers will only rarely take a write lock on the type tree, so this should have low contention.

This does require passing the type_tree along a bunch when evaluating `ParsedContentFilter`.

I also realized we don't have _any_ tests for event subscriptions, so I wrote a proper one with some filtering. No major issues there though, it worked fine out of the gate. There's probably still more we can do to make event filter building a little bit more ergonomic.

### DIFF
--- a/async-opcua-macros/src/events/gen.rs
+++ b/async-opcua-macros/src/events/gen.rs
@@ -156,6 +156,10 @@ pub(crate) fn generate_event_impls(event: EventStruct) -> syn::Result<TokenStrea
             fn time(&self) -> &opcua::types::DateTime {
                 self.base.time()
             }
+
+            fn event_type_id(&self) -> &opcua::types::NodeId {
+                self.base.event_type_id()
+            }
         }
 
         impl opcua::nodes::EventField for #ident {
--- a/async-opcua-nodes/src/events/evaluate.rs
+++ b/async-opcua-nodes/src/events/evaluate.rs
@@ -8,6 +8,8 @@ use opcua_types::{
     VariantScalarTypeId, VariantTypeId,
 };
 
+use crate::TypeTree;
+
 use super::{
     event::Event,
     validation::{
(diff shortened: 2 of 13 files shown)


lov3b (commented):
it looks good to me code wise. One thing that I though of is that you're passing true on the parameter `allow_complex_operators` to `validate_where_clause` in *validation.rs*. To me it looks like it will then go ahead and say that the input is valid for `InView`, `OfType` and `RelatedTo`, not just the implemented feature `OfType`. The snippet that I'm thinking of is:

```rust
if !allow_complex_operators
    && matches!(
        e.filter_operator,
        FilterOperator::InView | FilterOperator::OfType | FilterOperator::RelatedTo
        )
    {
    return (
        ContentFilterElementResult {
                status_code: StatusCode::BadFilterOperatorUnsupported,
                operand_status_codes: None,
                operand_diagnostic_infos: None,
                },
                None,
            );
        }
```

With other words: You're passing true to `allow_complex_operators` but you're not supporting them all, just OfType. Maybe the validation will wrongfully mark unimplemented operators as valid.

einarmo:
True, I suppose we should change that... We should probably just remove that parameter entirely and hard code which we support in the validation.

einarmo:
I changed it so we're explicitly excluding those. We may care about different filters having different allowed operators, for example for the `query` endpoint.

svanharmelen (approved):

PR #126, opened 2025-07-16
System for cheap node ID references

einarmo:
This started as a bit of an experiment, but I think it may be a decent change. (I am open for dissenting opinions, for sure...). An issue I sometimes run into is that in order to do stuff like look in the address space, notify the server of subscriptions, etc. you sometimes have to allocate a `NodeId`. This is annoying, and not really that elegant. I.e. you end up needing to do `&NodeId::new(0, "hello")`, which will allocate a `String`, even if that isn't really necessary, since we just need a reference.

`hashbrown`, the library we use for hashmaps, has a trait called `Equivalent`, which lets you use different types as keys in a hashmap, so long as you assert that they _hash_ identically.

This PR first splits the `node_id` module into smaller parts, because it is already way too large, then implements the system I described above as a combination of an `IdentifierRef` trait, a `NodeIdRef<T: IdentifierRef>` type, and an `IntoNodeIdRef<'a>` trait. I introduce this a bit in `References` and `AddressSpace`, mostly as a proof of concept. As a bonus, it is actually almost entirely backwards compatible. One breaking change I made is only partially unrelated, I removed the `From<i32>` implementation for `NodeId`, because we _really_ should avoid panicking `From` implementations, especially in such a critical place.

If this works out, I have an idea for a much easier to use and more flexible system for notifying the subscription cache of changes, which can come in a separate PR.

### DIFF
--- a/async-opcua-nodes/src/references.rs
+++ b/async-opcua-nodes/src/references.rs
@@ -1,5 +1,10 @@
+use std::hash::Hasher;
+
 use hashbrown::{Equivalent, HashMap, HashSet};
-use opcua_types::{BrowseDirection, NodeId};
+use opcua_types::{
+    node_id::{IdentifierRef, IntoNodeIdRef, NodeIdRef},
+    BrowseDirection, Identifier, NodeId,
+};
 
 use crate::{ImportedReference, ReferenceDirection, TypeTree};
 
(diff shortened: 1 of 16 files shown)


lov3b (approved):
I like the initiative. It's nice to avoid allocations, and it seems that the API is as usable as earlier with `impl IntoNodeIdRef<'a>`. I think that it looks good :p. Good job. Maybe we should add something to the documentation that states the relation between `NodeIdRef` and `NodeId`, and the motivation for them.

oroulet:
@einarmo I do not know enough the internal to really review that MR and the other open ones. But I can approve if you want/need

svanharmelen (approved):

PR #131, opened 2025-08-03
do not try to parse empty remote certificate

oroulet:
found out a server was returning

 "REMOTE CERT" = "REMOTE CERT"
[/home/olivier/dev/async-opcua/async-opcua-core/src/comms/secure_channel.rs:353:9] &remote_cert = ByteString {
    value: Some(
        [],
    ),
}

or in whireshark 
image

and the rust client was trying to parse that and failed with invalid certificate. That simple line change fixes it for me. but I am far from sure this is correct

### DIFF
--- a/async-opcua-core/src/comms/secure_channel.rs
+++ b/async-opcua-core/src/comms/secure_channel.rs
@@ -350,7 +350,7 @@ impl SecureChannel {
         &mut self,
         remote_cert: &ByteString,
     ) -> Result<(), StatusCode> {
-        self.remote_cert = if remote_cert.is_null() {
+        self.remote_cert = if remote_cert.is_null() || remote_cert.is_empty() {
             None
         } else {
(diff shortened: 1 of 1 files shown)


einarmo (approved):
This probably isn't harmful. I can't find any details in the standard. There may be a separate bug here, the standard states:

> If the [securityPolicyUri](https://reference.opcfoundation.org/search/17?t=securityPolicyUri) is NONE and none of the [UserTokenPolicies](https://reference.opcfoundation.org/search/17?t=UserTokenPolicies) requires encryption, the [Client](https://reference.opcfoundation.org/search/17?t=Client) shall ignore the [ApplicationInstanceCertificate](https://reference.opcfoundation.org/search/17?t=ApplicationInstanceCertificate).

It's possible we aren't properly doing this, or some other validation is missing.

PR #137, opened 2025-08-09
Forward fatal errors from subscription loop

einarmo:
If we get an error like `BadSessionClosed` or `BadSessionIdInvalid` from the subscription event loop, it _usually_ means that we are already in the process of closing the session.

In some cases it may, however, be the first indication of session failure, so we can forward it to the main session event loop and trigger session shutdown. This will reconnect according to normal rules.

The server would have to be _very_ misbehaved to respond to _anything_ with `BadSessionClosed` or `BadSessionIdInvalid` and disconnecting being the wrong response on our end.

### DIFF
--- a/async-opcua-client/src/session/services/subscriptions/event_loop.rs
+++ b/async-opcua-client/src/session/services/subscriptions/event_loop.rs
@@ -17,6 +17,11 @@ pub enum SubscriptionActivity {
     /// A publish request failed, either due to a timeout or an error.
     /// The publish request will typically be retried.
     PublishFailed(StatusCode),
+    /// Fatal failure, a publishing request has failed fatally in a way
+    /// that indicates it will not recover on its own.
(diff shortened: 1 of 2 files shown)


async-opcua-client/src/session/services/subscriptions/event_loop.rs
```diff
@@ -17,6 +17,11 @@ pub enum SubscriptionActivity {
     /// A publish request failed, either due to a timeout or an error.
     /// The publish request will typically be retried.
     PublishFailed(StatusCode),
+    /// Fatal failure, a publishing request has failed fatally in a way
+    /// that indicates it will not recover on its own.
+    /// This typically means the client has lost connection to the server.
+    /// When this is received by the session event loop it triggers a session restart.
```
svanharmelen, line 23: So does this change behavior for clients that also hook into the event_loop and have additional logic around reconnects?

PR #142, opened 2025-09-03
Use is-empty or is-null-or-empty in a few places

einarmo:
Some clients poorly distinguish between empty strings and null strings, this is a recurring problem. Prefer using `is_empty` for `UAString` and `is_null_or_empty` for `ByteString`. It's annoying that `is_empty` has different semantics for `UAString` and `ByteString`, but it would be a very ugly breaking change to align them now.


svanharmelen (commented):
Except for the Clippy comment, this looks good...

einarmo:
I fixed the clippy error in the other PR you approved.

svanharmelen (approved):

PR #152, opened 2025-09-15
Refactor building server connector

einarmo:
This is necessary if we want to let users swap out the connector. The final implementation is moderately hairy, because we connection establishement has several stages:

 - Fetch endpoint URLs from the server _or_ use an explicit endpoint.
 - Create a reusable TCP connector.
 - Connect to the server.

Because we want old methods on the clinet to keep working, I add layer in-between the new `ConnectionSource` trait, and the `Connector`, which is essentially just `TryInto<Arc<dyn Connector>>`, `ConnectorBuilder`.

This means that the flow is

 - `ConnectionSource` accepts an endpoint URL and produces a `ConnectorBuilder`.
 - `ConnectorBuilder` creates a `Connector`.
 - `Connector` is reused each time we need to re-connect to create a `TcpConnection`.

In the future we might even want to be able to create a different type of `Connection`, but I would want us to look into how we can ensure that is still efficient, i.e. avoid boxing futures too much.

Currently this doesn't change much at all. It changes the signature for the client connect methods in a backwards compatible way, and it makes it possible to define custom connectors, but the real purpose of this change is to be able to add support for `ReverseConnect`.

### DIFF
--- a/async-opcua-client/src/session/connection.rs
+++ b/async-opcua-client/src/session/connection.rs
@@ -1,19 +1,14 @@
 use std::{str::FromStr, sync::Arc};
 
+use crate::{
+    transport::{tcp::TransportConfiguration, Connector, ConnectorBuilder},
+    AsyncSecureChannel, ClientConfig, IdentityToken,
+};
 use opcua_core::{comms::url::is_opc_ua_binary_url, config::Config, sync::RwLock};
 use opcua_crypto::{CertificateStore, SecurityPolicy};
 use opcua_types::{
-    ContextOwned, EndpointDescription, MessageSecurityMode, NamespaceMap, NodeId, StatusCode,
-    TypeLoader, UserTokenType,
-};
-use tracing::error;
-
-use crate::{
-    transport::{
-        tcp::{TcpConnector, TransportConfiguration},
-        Connector,
-    },
-    AsyncSecureChannel, ClientConfig, IdentityToken,
+    ContextOwned, EndpointDescription, Error, MessageSecurityMode, NamespaceMap, NodeId,
+    StatusCode, TypeLoader, UserTokenType,
 };
 
 use super::{Client, EndpointInfo, Session, SessionEventLoop};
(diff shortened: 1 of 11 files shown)


async-opcua-client/src/session/connection.rs
```diff
@@ -1,42 +1,72 @@
 use std::{str::FromStr, sync::Arc};
 
+use crate::{
+    transport::{tcp::TransportConfiguration, Connector, ConnectorBuilder},
+    AsyncSecureChannel, ClientConfig, IdentityToken,
+};
```
svanharmelen, line 6: Why did you move the `use crate` block in between the `use std` and `use xxx (3rd party crates)` blocks? I think most Rust projects place the crates own packages last as do we in most (if not all) other files.

async-opcua-client/src/session/connection.rs
```diff
@@ -116,59 +147,84 @@ impl<T, R> SessionBuilder<'_, T, R> {
                 .is_some_and(|e| e.iter().any(|p| p.token_type == UserTokenType::IssuedToken)),
         }
     }
+
+    /// Set the connection source to use. This is used to create the transport
+    /// connector. Defaults to a direct TCP connection, implemented by `()`.
+    pub fn with_connector<C2>(self, connection_source: C2) -> SessionBuilder<'a, T, R, C2>
```
svanharmelen, line 153: Just a tiny little suggestion:

```suggestion
    pub fn with_connector<CS>(self, connection_source: CS) -> SessionBuilder<'a, T, R, CS>
```

async-opcua-client/src/transport/connect.rs
```diff
@@ -26,8 +26,50 @@ pub trait Connector: Send + Sync {
...
+
+    /// Get the default endpoint for this connector.
+    fn default_endpoint(&self) -> EndpointDescription;
+}
+
+/// Trait for types that can be used to create a connector.
+/// Implemented for `String`, `&str`, `&String`, and any type that implements the `Connector` trait.
+pub trait ConnectorBuilder: Send + Sync {
+    /// Create a new connector for the specific endpoint URL.
+    fn build(self) -> Result<Arc<dyn Connector + Send + Sync>, Error>;
```
svanharmelen, line 39: Maybe I just didn't look close enough at the PR, but why do we need the `Arc` here?

async-opcua-client/src/session/connection.rs
einarmo: I'll be honest. I spend zero brain-cycles on import ordering, just relying on rustfmt to deal with it. Interestingly I think rustfmt preserves import blocks if they have blank lines between them. What happened here is I removed a blank line, and it formatted `crate::` above `opcua*::`. Not sure why, really. I can revert it.

PR #154, opened 2025-09-19
Client support for reverse connect

einarmo:
This implements the OPC-UA reverse connection protocol in the client only, with a test using the .NET server.

While working with this I hit a panic in the secure channel, which we should avoid, so I fixed it. Generally, this is not a very complex implementation. To make it possible to provide a `:0` address, I let users provide the TCP listener to the reverse connect source, which works fine.

It's a little clunky to test this using the .NET server, so I hope to have a few more thorough tests once we have server support.


svanharmelen:
Good that I waited with testing this one 😏 Will checkout the branche today en give it a go...

svanharmelen (approved):
Looks good to me 👍🏻

PR #155, opened 2025-09-19
Extract the subscription event loop

einarmo:
Another slightly weird feature that I want for some stuff I'm working on. One thing I'm increasingly looking at is having ways to create "advanced" clients using the SDK without relying on the `Session` directly.

A barrier to that is the subscription event loop. It's challenging to implement yourself, and we will likely iterate on it in the future as well, meaning that implementations are likely to get out of sync without noticing.

This PR pulls most of the event loop out into a separate file using generics, so that users could define their own event loop, with other ways to make publish requests. This would also let us make small modifications to make the event loop more modular and configurable.

As a side-benefit, it also splits the very large subscription event loop function up a bit, and makes the state a bit more explict, which is probably a good thing.

### DIFF
--- a/async-opcua-client/src/lib.rs
+++ b/async-opcua-client/src/lib.rs
@@ -156,6 +156,11 @@ pub mod services {
         SetPublishingMode, SetTriggering, TransferSubscriptions, TranslateBrowsePaths,
         UnregisterNodes, Write,
     };
+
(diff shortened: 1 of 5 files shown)


svanharmelen (approved):
Looks good and works as expected 👍🏻

PR #166, opened 2025-10-28
Fix/default for method execution

bengeisler:
I'm unsure what to do about [these generated](https://github.com/FreeOpcUa/async-opcua/blob/master/async-opcua-types/src/generated/types/method_attributes.rs#L21-L22) `MethodAttributes` -- shouldn't `Default` return `true` for `executable` and `user_executable`?


bengeisler:
If we don't allow execution the server will return `Bad_UserAccessDenied`. Maybe a better option would be to return `Bad_NotSupported` for those?

einarmo (approved):

einarmo:
I'm fine with merging this as-is then, if you're happy.

PR #185, opened 2025-12-09
Fix busy loop

### LINKED ISSUE #184: Performance Regression
PR #155 introduced a Performance Regression by introducing a busy loop.
Within `EventLoopState::tick` the select branch `_ = self.wait_for_next_tick(next_publish) => ActivityOrNext::Next(next_publish)` leads to a busy loop because once the sleep future returns, it continues to create sleep futures which instantly resolve....
this whole branch seems unnecessary as it does no work at all...

jraffeiner:
Fixes #184 
by removing the unnecessary select branch and its parameters


einarmo:
There is a bug here, but I think the fix is to correctly update the next publish time, instead of just removing the loop entirely. I'll look into it. We'll probably have to release a patch version...

jraffeiner:
Yes #186 fixes this issue. It did seem weird for this branch to not do any work, strangely enough the client seemed to work just fine withouth it, but could be that it was some lucky coincidence...

PR #186, opened 2025-12-09
Fix subscription loop

einarmo:
This is https://github.com/FreeOpcUa/async-opcua/issues/184. I missed this entire part of the logic. This is severe enough that I'm bumping a version as well.

I'll look into writing a test for this in the future. The fix can be verified though, as the client currently has very noticable CPU usage that this removes.


svanharmelen (approved):
Ouch!

Issue #194, opened 2026-01-24
Expose a `connect_with_stream` or transport-connector API for the OPC UA client (tokio)


shiyuecamus:
### Background / context
I use the `opcua` (async-opcua) client APIs to:

- Discover endpoints
- Select endpoint by security policy/mode
- Connect and run the session event loop
- Perform `read` / `write` / subscriptions

In production we need:

- **Transport bytes metering** (bytes_in/out) for per-channel/per-device observability
- **Custom socket options** (bind local address/port, keepalive, timeouts, DSCP, etc.)
- **Custom TLS / mTLS** integration in some deployments
- **Deterministic testing** using in-memory streams or mocked I/O

Today the client creates/manages the underlying TCP/TLS transport internally (e.g. `connect_to_endpoint_directly(...)`). Downstream applications can’t wrap or inject the underlying I/O stream.

### Requested feature
Add an API that allows connecting a session using an existing stream, or providing a connector that creates the stream.

#### Option A (preferred): connect using an existing stream

```rust
// Pseudo-API sketch (exact types may differ in your codebase)
impl Client {
    pub async fn connect_to_endpoint_with_stream<S>(
        &mut self,
        endpoint: EndpointDescription,
        identity: IdentityToken,
        stream: S,
    ) -> Result<(Arc<Session>, SessionEventLoop), StatusCode>
    where
        S: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin + Send + 'static;
}
```

#### Option B: connector injection on `ClientBuilder`

```rust
// Pseudo-API sketch
ClientBuilder::transport_connector(|endpoint_url: &str| async move {
    // Create a TCP/TLS stream with custom options, then return it
    Ok(stream)
});
```

### Expected behavior / acceptance criteria
- Existing APIs (`connect_to_endpoint_directly`, etc.) remain unchanged.
- Injected stream path produces the same `(Session, SessionEventLoop)` pair.
- The OPC UA protocol/framing logic remains inside the crate; only transport creation is externalized.
- Works for both:
  - SecurityPolicy=None (plaintext)
  - Secure endpoints (if applicable in current architecture)

### Why this matters (benefits)
- Enables **first-class observability** (bytes_in/out, backpressure, reconnect patterns) without OS tooling.
- Enables enterprise transport constraints common in OT networks.
- Improves testability without brittle integration environments.

### Notes (our current usage)
[...]

PR #195, opened 2026-01-27
Make the client generic over transport type

einarmo:
This was a lot easier than I expected, which is why I think we should go ahead with it. It would allow us to solve #194 much more easily, and also open for other transports.

While I'm fine with `dyn` in the stack, I'd really prefer to avoid boxing our futures on the critical path. An alternative to this would be to box `Transport`. I thought using generics would be a lot messier, but it turns out that the `Session` doesn't actually need the connector at all, only the event loop does. This means that we can make `SessionEventLoop` generic over connector type, but keep `Session` completely without generics.

This is acceptable to me, since I expect most users just `spawn` the event loop and never pass it around manually. (If they do they'll have to make methods generic over `Connector`, probably, like I had to in the tests). Note that none of the sample applications had any breakage due to this.

This _technically_ already allows for external transport implementations, but it's pretty complicated, so if we're going to open up for that we probably want some layers of abstraction to make it easier to implement on the core parts of the transport. For #194 I was considering a `TcpStreamTransport<R: AsyncRead, W: AsyncWrite>`, or something along those lines, to replace the current `TcpTransport`.

There's also a separate commit with clippy lints for 1.93.

### DIFF
--- a/async-opcua-client/src/transport/tcp.rs
+++ b/async-opcua-client/src/transport/tcp.rs
@@ -181,6 +181,7 @@ impl TcpConnector {
 
 impl Connector for TcpConnector {
     type Transport = TcpTransport;
+
     async fn connect(
         &self,
         channel: Arc<SecureChannelState>,
@@ -411,6 +412,7 @@ impl ReverseTcpConnector {
 
 impl Connector for ReverseTcpConnector {
     type Transport = TcpTransport;
+
     async fn connect(
         &self,
         channel: Arc<SecureChannelState>,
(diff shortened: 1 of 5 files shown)


async-opcua-client/src/transport/tcp.rs
```diff
@@ -178,8 +179,8 @@ impl TcpConnector {
     }
 }
 
-#[async_trait]
 impl Connector for TcpConnector {
+    type Transport = TcpTransport;
```
svanharmelen: Here are a few more. Think these are the last ones.

PR #209, opened 2026-05-13
corrects the sample server config to match the `Limits` struct fields

bengeisler:
This PR corrects the `server.conf` to match the actual structure of the server config. Right now the incorrect keys were silently swallowed and replaced with the default values of the config. 

I also added a test that should prevent the config file going out of sync with the code.


einarmo (approved):
Neat, thanks!

Issue #217, opened 2026-05-26
Add support for dynamic structured datatypes for XML load (server)


OSjoerdWie:
While loading my nodeset into my custom server, I ran into issues with UAExpert, since no proper field derivation nor default binary encodings were set for dynamic structures/extension objects.

PR #218, opened 2026-05-26
Derive structure fields and binary encodings from the XML file.

OSjoerdWie:
This should fix issue #217

### DIFF
--- a/async-opcua-nodes/src/xml.rs
+++ b/async-opcua-nodes/src/xml.rs
@@ -1,15 +1,22 @@
 use std::{
-    path::Path, str::FromStr, sync::{Arc, OnceLock}
+    path::Path,
+    str::FromStr,
+    sync::{Arc, OnceLock},
 };
 
 use hashbrown::{HashMap, HashSet};
 use opcua_types::{
-    Context, DataTypeDefinition, DataTypeId, DataValue, DecodingOptions, EnumDefinition, EnumField, Error, LocalizedText, NodeClass, NodeId, QualifiedName, ReferenceTypeId, StructureDefinition, StructureField, StructureType, TypeLoader, TypeLoaderCollection, Variant
+    Context, DataTypeDefinition, DataTypeId, DataValue, DecodingOptions, EnumDefinition, EnumField,
+    Error, LocalizedText, NodeClass, NodeId, QualifiedName, ReferenceTypeId, StructureDefinition,
+    StructureField, StructureType, TypeLoader, TypeLoaderCollection, Variant,
 };
 use opcua_xml::{
-    XmlError, load_nodeset2_file, schema::ua_node_set::{
-        self, ArrayDimensions, ListOfReferences, UADataType, UAMethod, UANodeSet, UAObject, UAObjectType, UAReferenceType, UAVariable, UAVariableType, UAView
-    }
+    load_nodeset2_file,
+    schema::ua_node_set::{
+        self, ArrayDimensions, ListOfReferences, UADataType, UAMethod, UANodeSet, UAObject,
+        UAObjectType, UAReferenceType, UAVariable, UAVariableType, UAView,
+    },
+    XmlError,
 };
 use regex::Regex;
 use tracing::warn;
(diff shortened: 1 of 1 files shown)


async-opcua-nodes/src/xml.rs
```diff
@@ -239,25 +237,31 @@ impl NodeSet2Import {
...
+            let mut visited = HashSet::new();
+            let mut fields = self.collect_structure_fields(ctx, &base_data_type, &mut visited)?;
+            
+            // Add/override with this type's fields
             let mut any_optional = false;
-            let mut fields = Vec::with_capacity(def.fields.len());
             for field in &def.fields {
                 any_optional |= field.is_optional;
+                // Remove any base field with the same name (child overrides base)
+                fields.retain(|f| f.name.as_ref() != field.name);
```
einarmo: This makes me a little concerned, because the order of fields matters. I think this should probably be an index map, so we can track both the position of the field and keep its name unique.

This might actually be an issue in the types codegen as well, since I don't think we set it there either...

Really, we should be sharing more code between those two modules, but that's a larger project.

async-opcua-nodes/src/xml.rs
```diff
@@ -270,6 +274,77 @@ impl NodeSet2Import {
...
+            return Ok(all_fields);
+        }
+
+        // Prevent cycles
+        if !visited.insert(base_id.clone()) {
+            return Ok(all_fields);
+        }
+        
+        // Find the base DataType node in the loaded file
+        let base_node = self.file.nodes.iter().find_map(|n| {
```
einarmo: This is O(n^2) where n can be _quite_ large, which is also a little concerning. This is annoying to work around, especially if we want to keep the stream-like approach we're currently using... We'd have to pre-build a node map.

async-opcua-nodes/src/xml.rs
```diff
@@ -418,16 +493,39 @@ impl NodeSet2Import {
...
     }
 
-    fn make_data_type(&self, ctx: &Context<'_>, node: &UADataType) -> Result<ImportedItem, Error> {
+    fn make_data_type(&self, ctx: &Context<'_>, node: &UADataType, binary_encoding_types: &HashSet<String>) -> Result<ImportedItem, Error> {
         let base = self.make_base(ctx, &node.base.base, NodeClass::DataType)?;
+       
+        let mut base_data_type = NodeId::null();
+        let mut default_encoding = NodeId::null();
+        
+        let references = self.make_references(ctx, &base, &node.base.base.references)?;
```
einarmo, line 509: Notably this is unfortunately insufficient, since references may come from either side, meaning you need to consider _outgoing_ `HasSubtype` references of every other datatype node as well... It's part of what makes this so hairy. We lazily compute these in the codegen module:

https://github.com/FreeOpcUa/async-opcua/blob/f8d9cf82fd001d55c8c6bcff57ce42cfcc65b3ba/async-opcua-codegen/src/input/nodeset.rs#L211

einarmo (commented):
This is well done. It's clear that the dynamic nodeset import has fallen behind codegen. I'm concerned about the approach we use here, and about how this differs from the types codegen, where we already cover a lot of this functionality.

At the very least, since we do a first pass anyway, we should probably be pre-computing the stuff we need, like parent-type relationships and other encoding IDs, which is how we do it in codegen.

einarmo:
You'll want to run `cargo clippy` and `cargo fmt` on your code.

async-opcua-nodes/src/xml.rs
OSjoerdWie: Functionally, nothing has changed here. It was only moved upwards in order to do a pre-loop over the references. If I understand you correctly, you'd like to adjust the make_references? If so, should we bother diving into that topic in this PR?

async-opcua-nodes/src/xml.rs
OSjoerdWie: Good point. Had another look at the standard (https://reference.opcfoundation.org/specs/OPC-10000-3/8.48) which is clear on the order / occurrence of fields.

async-opcua-nodes/src/xml.rs
OSjoerdWie: This one should be fixed.

async-opcua-nodes/src/xml.rs
einarmo: My point was that technically, it's just as valid to put the `HasSubtype` relationship on the _parent_ as on the child, and some standards do this, I believe. (unfortunately).

That said, this is still an improvement. I think I want to pull a lot of this logic into async-opcua-xml, but that's a larger task. Ideally we'd share a lot of this between codegen and dynamic loading.

einarmo (approved):

PR #222, opened 2026-05-28
Ensure that the KeepAlive messages are not retransmitted.

### LINKED ISSUE #221: KeepAlive messages are retransmitted
When running my custom server, I notice the following in UAExpert:

Image

As it appears, the code queues the KeepAlive message for retransmission. However, according to the [standard](https://reference.opcfoundation.org/specs/OPC-10000-4/5.14.1#:~:text=Each%20keep%2Dalive,no%20Notifications.), KeepAlive messages are not `NotificationMessages`. Furthermore, the [standard ](https://reference.opcfoundation.org/specs/OPC-10000-4/5.14.1#:~:text=Sessions%20maintain%20a%20retransmission%20queue%20of%20sent%20NotificationMessages.) describes that only `NotificationMessages` should be maintained in a retransmission queue.

OSjoerdWie:
This PR should fix #221.

### DIFF
--- a/async-opcua-server/src/subscriptions/session_subscriptions.rs
+++ b/async-opcua-server/src/subscriptions/session_subscriptions.rs
@@ -532,6 +532,27 @@ impl SessionSubscriptions {
         }
     }
 
+    fn enqueue_retransmission_notification(
+        retransmission_queue: &mut VecDeque<NonAckedPublish>,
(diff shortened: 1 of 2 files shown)


einarmo (approved):
This is good, thanks! This part of the standard is particularly hard to parse, in my experience.

PR #223, opened 2026-05-28
Small fixes batch 2

oroulet:
More fixes discovered by Claude. They seem to make sense

### DIFF
--- a/async-opcua-client/src/transport/core.rs
+++ b/async-opcua-client/src/transport/core.rs
@@ -302,6 +302,11 @@ impl TransportState {
                 // right now. If we wait, we risk new messages using the new encryption keys arriving before
                 // we've updated the secure channel.
                 if let ResponseMessage::OpenSecureChannel(msg) = &message {
+                    let service_result = msg.response_header.service_result;
+                    if !service_result.is_good() {
+                        error!("OpenSecureChannel response failed, request_id = {req_id}: {service_result}");
+                        return Err(service_result);
+                    }
                     self.channel_state.end_issue_or_renew_secure_channel(msg).inspect_err(|e| {
                         error!("Failed to process OpenSecureChannel response, request_id = {req_id}: {e}");
                     })?;
--- a/async-opcua-client/src/transport/core.rs
+++ b/async-opcua-client/src/transport/core.rs
@@ -13,7 +13,7 @@ use opcua_core::comms::{
     chunker::Chunker, message_chunk::MessageChunk, message_chunk_info::ChunkInfo,
     tcp_codec::Message,
 };
-use opcua_types::{Error, StatusCode};
+use opcua_types::{DecodingOptions, Error, SimpleBinaryDecodable, StatusCode, UAString};
 
 use crate::transport::state::SecureChannelState;
 use crate::transport::RequestRecv;
@@ -210,11 +210,13 @@ impl TransportState {
     }
 
     fn process_chunk(&mut self, chunk: MessageChunk) -> Result<(), StatusCode> {
-        let secure_channel = trace_read_lock!(self.channel_state.secure_channel());
-        let chunk = secure_channel.verify_and_remove_security(chunk.data)?;
-
-        let chunk_info = chunk.chunk_info(&secure_channel)?;
-        drop(secure_channel);
+        let (chunk, chunk_info, decoding_options) = {
+            let secure_channel = trace_read_lock!(self.channel_state.secure_channel());
+            let chunk = secure_channel.verify_and_remove_security(chunk.data)?;
+            let chunk_info = chunk.chunk_info(&secure_channel)?;
+            let decoding_options = secure_channel.decoding_options();
+            (chunk, chunk_info, decoding_options)
+        };
         let req_id = chunk_info.sequence_header.request_id;
 
         self.sequence_numbers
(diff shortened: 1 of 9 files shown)


async-opcua-client/src/session/services/attributes.rs
```diff
@@ -597,14 +597,14 @@ impl Session {
...
-        Ok(HistoryRead::new(history_read_details, self)
+        HistoryRead::new(history_read_details, self)
             .timestamps_to_return(timestamps_to_return)
             .release_continuation_points(release_continuation_points)
             .nodes_to_read(nodes_to_read.to_vec())
             .send(&self.channel)
             .await?
             .results
-            .unwrap_or_default())
+            .ok_or(StatusCode::BadUnexpectedError)
```
einarmo: I don't necessarily agree with this, to be honest. OPC-UA servers do all sorts of weird stuff, and in other languages the distinction between None and Empty may be less obvious. As an OPC-UA SDK I think we should be permissive, to allow people to work around weird servers.

async-opcua-client/src/session/services/attributes.rs
```diff
@@ -625,12 +625,12 @@ impl Session {
...
         nodes_to_write: &[WriteValue],
     ) -> Result<Vec<StatusCode>, StatusCode> {
-        Ok(Write::new(self)
+        Write::new(self)
             .nodes_to_write(nodes_to_write.to_vec())
             .send(&self.channel)
             .await?
             .results
-            .unwrap_or_default())
+            .ok_or(StatusCode::BadUnexpectedError)
```
einarmo: Same here

async-opcua-client/src/session/services/attributes.rs
```diff
@@ -658,11 +658,11 @@ impl Session {
...
         history_update_details: &[HistoryUpdateAction],
     ) -> Result<Vec<HistoryUpdateResult>, StatusCode> {
-        Ok(HistoryUpdate::new(self)
+        HistoryUpdate::new(self)
             .details(history_update_details.to_vec())
             .send(&self.channel)
             .await?
             .results
-            .unwrap_or_default())
+            .ok_or(StatusCode::BadUnexpectedError)
```
einarmo: Same here

async-opcua-client/src/session/services/method.rs
```diff
@@ -135,12 +135,12 @@ impl Session {
...
         methods: Vec<CallMethodRequest>,
     ) -> Result<Vec<CallMethodResult>, StatusCode> {
-        Ok(Call::new(self)
+        Call::new(self)
             .methods_to_call(methods)
             .send(&self.channel)
             .await?
             .results
-            .unwrap_or_default())
+            .ok_or(StatusCode::BadUnexpectedError)
```
einarmo: Same here

async-opcua-client/src/transport/core.rs
```diff
@@ -390,3 +397,10 @@ impl TransportState {
         status
     }
 }
+
+fn decode_abort_status(chunk: &MessageChunk, chunk_info: &ChunkInfo) -> Option<StatusCode> {
```
einarmo: This should not be done this way. I found the thing the AI references, https://reference.opcfoundation.org/specs/OPC-10000-6/v1.05.07/6.7.3, which includes a string and an error code. We should be decoding this into a proper type, using the existing decoding framework.

I think we even should support sending these, sometimes, might be worth looking into.

einarmo (approved):

async-opcua-core/src/comms/sequence_number.rs
```diff
@@ -89,11 +93,14 @@ impl SequenceNumberHandle {
     pub fn validate_and_increment(&mut self, incoming_sequence_number: u32) -> Result<(), Error> {
         let expected = self.current();
         if incoming_sequence_number != expected {
-            // If the expected sequence number is the minimum value, and we are in legacy mode, then allow
-            // any value less than 1024.
-            // This is to handle the weird case in the OPC-UA standard stating that the
-            // first sequence number after we wrap around from the max can be any value less than 1024.
-            if self.is_legacy() && expected == self.min_value() && incoming_sequence_number < 1024 {
+            // Part 6 \xc2\xa76.7.2.4: after rolling over from MaxValue, the next sequence number
```
einarmo: Wait, this isn't true, is it? The standard doesn't explicitly state this, and I believe we had an issue with this at some point. The standard does in fact not say anything about what the _initial_ value of legacy sequence numbers is. Just

> SecurityPolicies with LegacySequenceNumbers set to TRUE, the SequenceNumber shall monotonically increase for all Messages and shall not wrap around until it is greater than 4 294 966 271 (UInt32.MaxValue – 1024). The first number after the wrap around shall be less than 1024.

Since we _can_ handle this, I'm not sure if it's worth it to make assumptions.

einarmo (changes requested):

async-opcua-core/src/comms/sequence_number.rs
einarmo: UA.NETStandard accepts _any_ sequence number here, which seems a little extreme? I suppose we should align with that, maybe?

```
if (m_firstReceivedSequenceNumber &&
    (
        !CryptoUtils.IsEccPolicy(SecurityPolicyUri) ||
        (CryptoUtils.IsEccPolicy(SecurityPolicyUri) && (sequenceNumber == 0))))
{
    m_remoteSequenceNumber = sequenceNumber;
    m_firstReceivedSequenceNumber = false;
    return true;
}
```

oroulet: I will remove that commit. I do not see any reason not to accept anything

oroulet: commit entirely removed

einarmo (approved):
Looks good! I have some upcoming work that changes some of these things as well, and this made me think we should probably be sending FinalError type messages in a few cases where I think we just cut the connection now. Worth looking into, at least.

PR #224, opened 2026-05-31
Client error improvements

einarmo:
In general, I strongly prefer using `Error` where possible, for a few reasons:
    
 - It lets us pass structured context. OPC-UA status codes can be very opaque.
 - It's extensible. It's very possible we'll be able to include some clever extensions in the future, like opt-in stack-traces which is _very useful for debugging.

This makes more or less everything in the client return `Error`, which is somewhat breaking. Note that Error implements `Into<StatusCode>`, so clients using `?` to return `StatusCode` will continue to work. I do expect some breakage, however. It really should not be too difficult to fix for most users.


oroulet:
makes a lot of sense..

oroulet (approved):
did not look into details but principle is very good

PR #226, opened 2026-05-31
Automate releases

einarmo:
Releases are triggered by just updating the version of all crates. Right now we also require them to be the same. To be safe, I made it so that releases run in a special "CD" environment, that is restricted to master and requires approval.


oroulet (approved):
I do not know the github specific things but here again the idea is good

PR #228, opened 2026-06-01
Set is_for_events to true for history_read_events.

OSjoerdWie:
In `history_read_events` the call to `self.validate_history_read_nodes` should be done with `true` for the `is_for_events` parameter instead of `false`.


einarmo (approved):
Oops! Don't think we have tests for all history read variants, though we should. Thanks!

PR #229, opened 2026-06-11
Fix ReferenceType::from_attributes setting symmetric from is_abstract

occamsshavingkit:
`ReferenceType::from_attributes` copies `attributes.is_abstract` into the `symmetric` attribute when the SYMMETRIC mask bit is set, so any ReferenceType node created through the AddNodes service gets a wrong `Symmetric` value whenever it differs from `IsAbstract`.

One-line fix: use `attributes.symmetric`.

Found while reviewing the downstream fork at occamsshavingkit/async-opcua; the crate's existing tests pass with the change.


einarmo (approved):
Thanks!

PR #230, opened 2026-06-12
fix(server): reclaim connection_map slot when a connection task panics

occamsshavingkit:
## Problem

When a spawned connection task panics, `tokio::spawn` resolves the `JoinHandle` as `Err(JoinError)`. The select loop logs the error but **does not call `self.connection_map.remove(&id)`** because the panic path has no counter to match on:

```rust
Err(e) => error!("Connection panic! {e}")  // slot never freed
```

The connection counter slot stays in `connection_map` forever, permanently consuming one `max_connections` slot. Under sustained load with buggy node manager code, this eventually exhausts the connection limit and the server stops accepting new connections.

## Fix

Wrap both spawn sites (accept and reverse connect) with `AssertUnwindSafe` + `catch_unwind` so the task always returns its counter whether it exits normally or panics. The `Ok(id)` path in the select loop then handles cleanup in both cases.

A `log_connection_panic` helper extracts the message from the panic payload (covers both `&str` and `String` payloads).

## Testing

Verified with `cargo check -p async-opcua-server`. The existing `connection_limit_tests` suite covers the slot-reclamation path.

### DIFF
--- a/async-opcua-server/src/server.rs
+++ b/async-opcua-server/src/server.rs
@@ -53,6 +53,15 @@ struct ConnectionInfo {
     command_send: tokio::sync::mpsc::Sender<ControllerCommand>,
 }
 
+fn log_connection_panic(id: u32, payload: Box<dyn std::any::Any + Send>) {
+    let message = payload
+        .downcast_ref::<&str>()
+        .copied()
+        .or_else(|| payload.downcast_ref::<String>().map(String::as_str))
+        .unwrap_or("unknown panic payload");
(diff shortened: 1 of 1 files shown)


einarmo (approved):
Fair enough, this is a bug. I do think we should probably do a little cleanup here, there's some code duplication that's been a bit stubborn. _really_ we should also be able to just associate a handle with the connection counter, so we don't have to do this, but it might get more noisy than is useful.

PR #231, opened 2026-06-12
perf(core): allocation-free transmit path via streaming chunk encoder

occamsshavingkit:
## Summary

Two related changes that together eliminate per-message heap allocations from the OPC UA transmit path.

### 1. `MessageChunk::data: Vec<u8>` → `bytes::Bytes`

`bytes::Bytes` is reference-counted, cheaply cloneable, and slice-shareable. Changes:
- `MessageChunk` gets `Clone` (required by the streaming encoder).
- `SimpleBinaryDecodable` wraps the read buffer in `Bytes::from`.
- `verify_and_remove_security` / `verify_and_remove_security_server` accept `Bytes` directly — zero-copy at the receive boundary. Decrypt paths that need mutation still do a single `to_vec()` inside.
- Construction sites in `secure_channel.rs` / `client/transport/core.rs` use `.into()`.

### 2. Streaming `ChunkingStream` encoder in `SendBuffer`

`Chunker::encode` previously allocated a `Vec<u8>` body per chunk, then `MessageChunk::new` copied it into another per-chunk `Vec`. The new path:

- `ChunkingStream` writes full chunks (header + body) directly into a caller-provided `BytesMut` buffer owned by `SendBuffer`.
- Each chunk is emitted as a zero-copy `Bytes` slice into the output `Vec<MessageChunk>`.
- `SendBuffer::reserve()` resets the `BytesMut` position so the storage is reused for the next message — **steady-state transmit encoding allocates nothing**.
- Chunk body sizes are computed up front, so the header is written before the body streams in, producing byte-identical output to the old path (verified by existing crypto tests).
- `Chunker::encode` is kept as a compatibility shim over `encode_into`.
- Adds a regression test asserting `BytesMut` capacity is stable (not reallocated) across messages.

## Testing

`cargo check` clean on core, server, and client crates. Existing chunker / secure-channel / crypto tests cover correctness; the new `send_buffer_storage_reuse` test covers the allocation-free invariant.

### DIFF
--- a/async-opcua-core/src/comms/chunker.rs
+++ b/async-opcua-core/src/comms/chunker.rs
@@ -18,11 +18,11 @@ use crate::{
 use opcua_crypto::SecurityPolicy;
 use opcua_types::{
     encoding::BinaryEncodable, node_id::NodeId, status_code::StatusCode, BinaryDecodable,
-    EncodingResult, Error, ObjectId,
+    EncodingResult, Error, ObjectId, SimpleBinaryEncodable,
 };
 use tracing::{debug, error, trace};
 
-use super::message_chunk::MessageChunkType;
+use super::{message_chunk::MessageChunkType, security_header::SequenceHeader};
 
 /// Read implementation for a sequence of message chunks.
 /// This lets us avoid allocating a buffer for the message.
@@ -102,13 +102,23 @@ impl<'a, T: Iterator<Item = &'a MessageChunk>> Read for ReceiveStream<'a, T> {
     }
 }
 
-struct ChunkingStream<'a> {
+/// Streaming chunk encoder that writes complete chunks (headers + body)
+/// contiguously into a caller-provided [`BytesMut`], emitting each chunk as a
+/// zero-copy [`Bytes`](bytes::Bytes) slice of that storage. With a reused
+/// connection-local storage buffer this makes steady-state transmit encoding
+/// allocation free: once the chunks from the previous message are dropped the
+/// storage allocation is reclaimed by `reserve`.
+struct ChunkingStream<'a, 'b> {
     secure_channel: &'a SecureChannel,
-    chunks: Vec<MessageChunk>,
+    storage: &'b mut bytes::BytesMut,
+    out: &'b mut Vec<MessageChunk>,
     expected_chunk_count: usize,
+    chunks_emitted: usize,
     max_body_per_chunk: usize,
-    next_buf: Vec<u8>,
-    buf_position: usize,
+    header_size: usize,
+    current_body_target: usize,
+    body_written: usize,
+    chunk_started: bool,
     is_closed: bool,
     sequence_number: SequenceNumberHandle,
     request_id: u32,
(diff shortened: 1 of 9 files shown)


async-opcua-core/src/comms/chunker.rs
```diff
@@ -142,112 +155,145 @@ impl<'a> ChunkingStream<'a> {
...
-                max_body_per_chunk,
-                next_buf: vec![0; next_buf_size],
-                buf_position: 0,
-                is_closed: false,
-                sequence_number,
-                request_id,
-                message_type,
-                message_size,
-            })
+            (message_size / max_body_per_chunk + 1, max_body_per_chunk)
```
einarmo: Is there a corner case here where message_size == max_body_per_chunk? You'd end up with a final chunk with size 0, right? I think we should maybe avoid that, since it could cause issues for some third party implementations. Not sure what the standard says about it.

async-opcua-core/src/comms/secure_channel.rs
```diff
@@ -872,12 +872,12 @@ impl SecureChannel {
...
         if message_header.message_type.is_open_secure_channel() {
             let (decrypted_chunk, _) =
-                self.decrypt_open_secure_channel(src, security_header, encrypted_range)?;
+                self.decrypt_open_secure_channel(src.to_vec(), security_header, encrypted_range)?;
             Ok(decrypted_chunk)
         } else if self.is_secure_connection() {
             let signature_size = self.security_policy.symmetric_signature_size();
             let signed_range = 0..(message_size - signature_size);
-            self.decrypt_chunk(src, security_header, signed_range, encrypted_range)
+            self.decrypt_chunk(src.to_vec(), security_header, signed_range, encrypted_range)
```
einarmo: From what I can tell, these methods only ever use a slice, or convert it back into a `Bytes`, so I think we can change them to take `Bytes` as well, unless I'm missing something. Right now this nice improvement to remove an allocation is mostly pointless since it's all lost right here where we do one extra allocation anyway.

einarmo (commented):
Not bad! Thanks! I have a few comments, and I'd also like you to rebase if possible, so that each commit builds. I prefer a bisectable history.

occamsshavingkit:
Thanks for the review! Addressed both points:

1. **Zero-body final chunk**: Replaced `message_size / max_body_per_chunk + 1` with `div_ceil(...).max(1)`, and the last-chunk body target now uses `message_size - chunks_emitted * max_body_per_chunk` so it's always correct regardless of divisibility.

2. **`to_vec()` allocations**: Changed `decrypt_open_secure_channel` and `decrypt_chunk` to accept `bytes::Bytes` directly. Both helpers already allocate their own output buffer internally, so the signature change is zero-cost.

Also rebased into two bisectable commits as requested — the `MessageChunk.data: Bytes` migration compiles standalone, and the chunker sits on top of it.

async-opcua-core/src/comms/secure_channel.rs
occamsshavingkit: I will work on these. I made a fork and went really far in implementing more of the OPC-UA standard, and now I am trying to figure out how to contribute everything back in reviewable chunks.

einarmo:
Lock file is out of date and contains `async-opcua-safety`, which is not one of ours.

occamsshavingkit (commented):
Whoops, good catch! Since we originally ran the `cargo` commands inside our massive fork (which contains our PubSub system and crates like `async-opcua-safety`), cargo dutifully added them to the `Cargo.lock` file. I've regenerated the lockfile against the pure upstream workspace, so it's perfectly clean now!

einarmo (approved):
Alright, this looks good, and is a reasonable improvement. In general I do think using `Bytes` / `BytesMut` is a good thing, since it is optimized for this kind of use.

PR #232, opened 2026-06-12
feat(client): SessionDropGuard for RAII session lifecycle (#143)

### LINKED ISSUE #143: impl Drop for Session
Hi all! Does anyone know if [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html) is purposefully not implemented for Session? If not, would people be open to having it implemented? It would be really helpful for RAII patterns and any sort of event loops. 

For implementation, we could maybe do something similar to what sqlx [does](https://docs.rs/sqlx-core_wasi/0.6.2/src/sqlx_core/pool/connection.rs.html#133) and just spawn the [disconnect](https://docs.rs/async-opcua-client/0.16.0/opcua_client/struct.Session.html#method.disconnect) function as a separate top level task that can be optionally be awaited if needed. 

I am happy to open a PR if folks are good with it. Thanks :)

occamsshavingkit:
## Summary

Closes #143.

Adds a `SessionDropGuard` that initiates a graceful server-side disconnect when dropped, enabling RAII-style session management without requiring an explicit `disconnect()` call.

### Design

- A `tokio::sync::watch` channel (sender in `Session`, receiver in `SessionEventLoop`) carries the close signal.
- `Session::close_on_drop()` returns a `SessionDropGuard` wrapping `Arc<Session>`.
- `Drop for SessionDropGuard` sends `true` on the watch channel.
- The event loop checks the channel at the top of every poll iteration (`borrow_and_update`) to handle the Disconnected/Connecting states, and adds a `close_rx.changed()` arm to the Connected-state `tokio::select!` for prompt response mid-connection.
- When the signal fires in the Connected state, `disconnect_inner(delete_subscriptions: true, …)` is queued just like a keep-alive failure.

### Usage

```rust
let (session, event_loop) = client.connect_to_matching_endpoint(endpoint, token).await?;
let _guard = session.close_on_drop(); // drop to disconnect
let handle = event_loop.spawn();
session.wait_for_connection().await;
// … use session …
drop(_guard); // sends CloseSession to server, event loop exits cleanly
handle.await;
```

### Notes

- `SessionDropGuard` is `#[must_use]` to prevent accidental immediate drop.
- `arc()` method gives access to the underlying `Arc<Session>` for sharing across tasks.
- The existing `disconnect()` / `disconnect_without_delete_subscriptions()` methods continue to work and are unaffected.

## Test plan

- [ ] Existing client test suite passes (`cargo test -p async-opcua-client`)
- [ ] Manual test: drop guard before event loop exits, verify server-side session is cleaned up
- [ ] `cargo clippy --all` clean

### DIFF
--- a/async-opcua-client/src/lib.rs
+++ b/async-opcua-client/src/lib.rs
@@ -127,8 +127,8 @@ pub use session::{
     Client, ConnectionSource, DataChangeCallback, DefaultRetryPolicy, DirectConnectionSource,
     EventCallback, HistoryReadAction, HistoryUpdateAction, MonitoredItem, MonitoredItemMap,
     OnSubscriptionNotification, OnSubscriptionNotificationCore, RequestRetryPolicy, Session,
-    SessionActivity, SessionBuilder, SessionConnectMode, SessionEventLoop, SessionPollResult,
-    Subscription, SubscriptionActivity, SubscriptionCallbacks, UARequest,
+    SessionActivity, SessionBuilder, SessionConnectMode, SessionDropGuard, SessionEventLoop,
+    SessionPollResult, Subscription, SubscriptionActivity, SubscriptionCallbacks, UARequest,
 };
 pub use transport::AsyncSecureChannel;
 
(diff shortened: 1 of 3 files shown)


einarmo (approved):

einarmo:
Pretty reasonable way to do this, IMO. We may still want to look into dropping the connection when the last session is dropped, but that's considerably harder, and this is non-intrusive and would solve the issue for any user with this need. Thanks!

PR #233, opened 2026-06-16
chore: CI fixes, clippy cleanups, and generated code sync

occamsshavingkit:
This PR backports a batch of minor chores, CI fixes, and code generation syncs from our fork. These are standalone, safe changes that do not rely on any structural deviations and compile cleanly against `master`.

### Changes Included
- **CI / Warnings**: Fixes for newer `clippy` lints, including generated code cleanups (e.g., `byte_len`) and `libelf` for TSN builds.
- **Codegen**: Syncs the generated code with the current `codegen` output, resolving any drift.
- **Debris Cleanup**: Removes accidentally tracked runtime logs and ignores `*.log` in `.gitignore`.


.github/workflows/main.yml
```diff
@@ -23,6 +23,10 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: ${{ matrix.toolchain }}
+      # libelf is needed by libbpf-sys, pulled in by xsk-rs when the
+      # pubsub tsn feature is enabled via --all-features.
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y libelf-dev
```
einarmo: This seems to refer to a feature that doesn't exist (yet) in this library.

einarmo (commented):
I understand that these are part of your fork, but I don't think these changes make sense in a vacuum. The clippy lint, _maybe_, though clippy should ignore the output of macro expansions, usually. As for the gitignore changes, I would prefer that we avoid adding things to the gitignore that aren't directly, by default, build-related. If you have loose files that annoy you, the appropriate way to deal with those is to add them to your global gitignore.

occamsshavingkit:
Good catch! That `libelf` dependency was accidentally ported over from our local fork where we're working on UDP/TSN PubSub features. Since that feature isn't upstreamed yet, installing it in CI here is totally useless. I've just pushed a commit removing it.

occamsshavingkit (commented):
Fair point about the `.gitignore`! I've reverted those additions to keep the repository configuration clean. (I'll handle my loose logs in my global gitignore as you suggested.)

einarmo (approved):
This is fine. I'll be squashing this into one commit, since it is now just a tiny fix.

PR #234, opened 2026-06-16
fix(types): return errors instead of panicking on subtyped-values struct kinds

occamsshavingkit:
This PR is a standalone backport from our fork that prevents panics when decoding subtyped-values struct kinds. 

It fixes the types implementation to return a proper `opcua_types::Error` rather than panicking. It includes a fuzzing target that was used to catch and verify this boundary condition.


einarmo (approved):
This looks reasonable, and the fuzz is nice to have, thanks.

PR #235, opened 2026-06-16
fix: make the workspace warning-free under CI

occamsshavingkit:
CI sets `RUSTFLAGS=-D warnings` which caused newer clippy versions to fail on accumulated warnings. This PR fixes all warnings across the workspace without resorting to `#[allow(...)]` attributes in generated code, ensuring a clean and reliable build.

- Replaced obsolete `for` loop patterns and needless borrows
- Fixed unidiomatic uses of `assert_eq!` with booleans
- Added missing indentations in doc comments
- Scoped redundant matches and returns


einarmo (approved):
Looks good, thanks.

PR #239, opened 2026-06-24
test(tier-a): lock in multi-AI cross-check Tier A findings

occamsshavingkit:
Completes **Tier A** of the multi-AI test-coverage cross-check (`specs/multi-ai-test-suites/UNIFIED-PROTOCOL.md`) — the four "potential real bug" items proposed independently by Antigravity/Codex. Each was probed against the server code first; **none was a crash or auth bypass**, so these lock the behaviour in as regression tests.

| Item | Claim | Finding |
|---|---|---|
| A1 | PercentDeadband caches EURange at create, never refreshes | **Confirmed** — but deliberate (avoids a node read per sample). Documented at the `eu_range` field; no live-refresh fix (would be a redesign). |
| A2 | MonitoredItem on a DeleteNodes'd node panics | **No panic** — sampler/read paths return graceful status. Test asserts server survives. |
| A3 | Reference cycle → recursive Browse/Translate stack overflow | **Not recursive** — TranslateBrowsePaths is iterative, bounded by the requested path length. Test pins this (cycle + 256-element path terminates). |
| A4 | Identity-token edges allow bypass | **Fail-closed** — empty-password UserName token rejected (distinct from the existing wrong-password case). |

New file: `async-opcua/tests/integration/tier_a.rs` (3 tests). One doc comment added at the `eu_range` field for A1.

All three tests pass locally; clippy + fmt clean.


occamsshavingkit:
Opened against the wrong remote — retargeting to my fork. Apologies for the noise.

PR #243, opened 2026-07-01
[codex] Remove hot-path locks with snapshots

occamsshavingkit:
## Summary

- add Spec Kit feature 046 for lock-removal/snapshot concurrency work
- implement TypeTree immutable snapshots for server hot-path metadata reads
- move response-size limit state onto SecureChannel instead of a global map
- add OPC UA clause-grounded focused tests and P3 gate evidence for subscription routing, PubSub config/cache, SQLite history, and SecureChannel renewal follow-ups

## Performance note

This PR does not prove a throughput improvement. Prior scratch samples from Feature 045 exist and are now recorded as candidate before-data, but the 046 after samples were collected with a different build/mode shape. Raw comparison against those prior samples shows a large throughput drop, so SC-004 remains inconclusive until before/after samples are captured apples-to-apples.

## Validation

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --all-features --locked -- -W clippy::await_holding_lock -W clippy::await_holding_refcell_ref`
- `cargo test --workspace --all-targets --all-features --locked`

Expected negative-path ERROR logs and `Gnuplot not found` bench messages appeared during the workspace test run, but the command exited 0.


occamsshavingkit:
Opened against upstream by mistake. Closing and reopening against occamsshavingkit/async-opcua.

PR #244, opened 2026-07-01
fix(limits): make two orphaned limit constants the source of truth

occamsshavingkit:
## Summary

Two `pub const` limits in `async-opcua-server::constants` advertised default values the server **never actually used** — the corresponding `defaults::*` functions hardcoded *different* literals:

| Constant | Declared | Real default (hardcoded in `defaults::`) |
|---|---|---|
| `DEFAULT_MAX_MONITORED_ITEMS_PER_SUB` (`lib.rs`) | `1000` | `10_000` |
| `MAX_NOTIFICATIONS_PER_PUBLISH` (`lib.rs`) | `0` | `1_000` |

Neither constant was referenced anywhere outside its own declaration, so nothing in-tree caught the drift — but they are `pub`, so any downstream caller importing them would get a value the server contradicts (off by 10×, or "unlimited" vs bounded). The drift dates to `4edcd63a4 perf(limits): embedded-sane footprint defaults`, which raised the defaults via literals and left the old constants behind.

## Fix

Make each constant the single source of truth: set it to the value the server actually uses, and have the `defaults::*` function return the constant.

- **Zero behavior change** — the compiled defaults were already `10_000` / `1_000` and remain so; the constants now *agree* instead of drifting.
- `0` remains available as an explicit "unlimited" opt-out for notifications-per-publish, now documented on the constant itself.
- Future edits happen in one place and can't drift again.

## Context

Found during an exhaustive audit of the server's limit machinery (all 41 `constants` values traced `const → defaults → config field → enforcement → ServerBuilder override`). Every other limit is correctly wired; these two orphaned constants were the only dead paths. The earlier "monitored-items-per-sub bypassable" concern is separately confirmed fixed (default is now non-zero and enforced).

## Verification

- `cargo test -p async-opcua-server --lib config::limits` — 7/7 pass (incl. bounded/non-zero default assertions and the field-name parity test)
- `cargo clippy -p async-opcua-server --all-features` — clean
- `cargo fmt --all` — clean


occamsshavingkit:
Opened against the wrong repository by mistake; recreating on the correct fork.

PR #246, opened 2026-07-03
Make sure to set `index_map` when loading imports

### LINKED ISSUE #245: NodeSet2Import does do not keep the `index_map` in mind when importing
This specific problem caused importing to crash the process. In our case this caused importing to fail on a NodeSet, because the nodeset the new namespace at index 1, while on the address space it was index 3. The attached pull request should provide some more clarity if necessary.

tetanw:
Should fix #245. I took the liberty to also correct the "Adding new namespace" to report the namespace as it will be in the address space instead of as if it is in the NodeSet.


einarmo (approved):
Looks good! Thanks.

PR #250, opened 2026-07-18
Release version 0.19.0

einarmo:
It's about time we do another release. This also serves as a test for the automated release process, so fingers crossed it works well.

I also cleaned up the docs in a few places that contained outdated or incorrect information.

The changelog is not quite exhaustive, but I picked the commits that I felt were notable. We auto-generate a full changelog in the release as well.


svanharmelen (approved):
🎉

PR #254, opened 2026-07-18
fix(types): bound Argument.value_rank against max_array_length

occamsshavingkit:
## Summary

Pre-authentication remote-DoS: `Argument::decode` did `array_dimensions.resize(value_rank as usize, 0)` with an attacker-controlled `i32 value_rank` read directly off the wire. A 30-byte crafted message with `value_rank ≈ 2.1e9` drives an ~8.45 GB allocation — reachable before any session/auth check, since `Argument` decodes as part of `Variant` and a `None`-security endpoint has no cryptographic gate in front of it.

Fix bounds `value_rank` by the connection's own negotiated `max_array_length` before the resize; ordinary `value_rank`s are unaffected.

Found via fuzzing on a downstream fork of this project ([occamsshavingkit/async-opcua](https://github.com/occamsshavingkit/async-opcua)). This patch was built and verified independently against this repository's own `master`, not carried over wholesale from the fork (which has diverged substantially). Privately disclosed to @einarmo ahead of this PR, at his request.

## Test plan

- [x] New regression tests decode the exact crafted artifact (and a minimal equivalent) and assert a clean decode error, not an allocation attempt
- [x] Existing `argument_decode_pads_dimensions_when_shorter_than_rank` and friends still pass — small, in-range `value_rank`s are unaffected
- [x] `cargo test -p async-opcua-types` — 0 failures
- [x] `cargo clippy -p async-opcua-types --all-targets` — clean

### DIFF
--- a/async-opcua-types/src/argument.rs
+++ b/async-opcua-types/src/argument.rs
@@ -112,6 +112,12 @@ impl BinaryDecodable for Argument {
         if value_rank > 0 {
             if let Some(array_dimensions) = array_dimensions.as_mut() {
                 let expected_rank = value_rank as usize;
+                if expected_rank > ctx.options().max_array_length {
+                    return Err(Error::decoding(format!(
+                        "Argument value_rank {value_rank} exceeds array length limit {}",
+                        ctx.options().max_array_length
+                    )));
+                }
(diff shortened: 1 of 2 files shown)


einarmo (approved):
Looks good, thanks.

PR #255, opened 2026-07-18
fix(core): enforce max_message_size on incoming TCP message headers

occamsshavingkit:
## Summary

Unauthenticated remote DoS surface: `MessageHeader::decode` read the declared `message_size` off the wire but never checked it against the connection's own negotiated `max_message_size` (the `DecodingOptions` parameter was received but ignored, per the leading underscore). This check runs before `OpenSecureChannel`, so it's reachable by any peer that's opened a raw TCP connection, no authentication required.

Reject `message_size > max_message_size` with `Bad_TcpMessageTooLarge` before the rest of the frame is buffered. `max_message_size == 0` continues to mean "unlimited," matching this repo's existing convention for that field.

Found and fixed on a downstream fork of this project ([occamsshavingkit/async-opcua](https://github.com/occamsshavingkit/async-opcua)). This patch was built and verified independently against this repository's own `master`, not carried over wholesale from the fork (which has diverged substantially). Privately disclosed to @einarmo ahead of this PR, at his request.

## Test plan

- [x] New regression tests: rejects a size above the max, accepts a size exactly at the max, and confirms `max_message_size = 0` still means unlimited
- [x] `cargo test -p async-opcua-core` — 0 failures
- [x] `cargo clippy -p async-opcua-core --all-targets` — clean

### DIFF
--- a/async-opcua-core/src/comms/tcp_types.rs
+++ b/async-opcua-core/src/comms/tcp_types.rs
@@ -95,10 +95,28 @@ impl SimpleBinaryEncodable for MessageHeader {
 }
 
 impl SimpleBinaryDecodable for MessageHeader {
-    fn decode<S: Read + ?Sized>(stream: &mut S, _: &DecodingOptions) -> EncodingResult<Self> {
+    fn decode<S: Read + ?Sized>(
+        stream: &mut S,
+        decoding_options: &DecodingOptions,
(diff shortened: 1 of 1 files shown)


einarmo (approved):
I _think_ we now check this somewhere else as well, but this is a good enough place to check it.

Thanks!

PR #256, opened 2026-07-18
fix(server): re-check session nonce before committing ActivateSession

occamsshavingkit:
## Summary

`activate_session` snapshots the session nonce under a read lock, then awaits `authenticate_endpoint()` with no lock held on the session, then commits `session.activate()` under a write lock without re-checking that the nonce is still current.

If a second, concurrent `ActivateSession` call for the same session completes in that window, it rotates the nonce. The first call then still commits its own activation against the stale, already-superseded nonce it captured earlier — weakening the replay/freshness guarantee the session nonce exists to provide.

This adds a fail-closed re-check under the same write lock that commits the activation: if the session's nonce has changed since the pre-auth snapshot, return `Bad_NonceInvalid` and mutate nothing. The check is extracted into a small, directly-testable helper (`nonce_still_current`) rather than inlined, since the interesting behavior is "does the comparison correctly detect a rotation," not the concurrency around it.

Found and fixed on a downstream fork of this project ([occamsshavingkit/async-opcua](https://github.com/occamsshavingkit/async-opcua)), where it was originally caught with a full concurrent-race integration test built against that fork's own test harness. This PR carries a smaller, focused unit test of the extracted check itself instead of porting that harness — happy to add a fuller concurrency test here too if maintainers want it, just didn't want to import fork-specific test scaffolding wholesale. The fix and test here were built and verified independently against this repository's own `master`.

Privately disclosed to @einarmo ahead of this PR, at his request.

## Test plan

- [x] New unit test constructs a `Session`, confirms the nonce check passes against its own current nonce, simulates an intervening rotation via `Session::activate`, and confirms the check now correctly reports the pre-rotation nonce as stale
- [x] `cargo test -p async-opcua-server` — 0 failures
- [x] `cargo clippy -p async-opcua-server --all-targets` — clean

### DIFF
--- a/async-opcua-server/src/session/manager.rs
+++ b/async-opcua-server/src/session/manager.rs
@@ -15,9 +15,9 @@ use tracing::{error, info};
 
 use crate::{identity_token::IdentityToken, info::ServerInfo};
 use opcua_types::{
-    ActivateSessionRequest, ActivateSessionResponse, CloseSessionRequest, CloseSessionResponse,
-    CreateSessionRequest, CreateSessionResponse, Error, NodeId, ResponseHeader, SignatureData,
-    StatusCode,
+    ActivateSessionRequest, ActivateSessionResponse, ByteString, CloseSessionRequest,
+    CloseSessionResponse, CreateSessionRequest, CreateSessionResponse, Error, NodeId,
+    ResponseHeader, SignatureData, StatusCode,
 };
 
 use super::{instance::Session, message_handler::MessageHandler};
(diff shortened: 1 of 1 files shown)


einarmo:
I need to dig a bit more, but I'm a little uncertain about this. It feels somewhat like a band-aid fix, while fundamentally we don't want users to ever perform _concurrent_ operations that may modify the secure channel state. There may be other similar issues (vulnerabilities? Not sure how you'd exploit this) elsewhere.

einarmo (approved):
I've dug into this a bit, and I think this solution is fine. I was considering a lock or lease based solution, so we can prevent this before beginning the activation at all, but this is _probably_ good enough. It's certainly quite simple.

Issue #257, opened 2026-08-02
rsa dependency (RUSTSEC-2023-0071) blocks safe use of RSA-based security policies


EvanL1:
## Summary

`async-opcua-crypto` depends on the RustCrypto `rsa` crate, which is affected by
[RUSTSEC-2023-0071](https://rustsec.org/advisories/RUSTSEC-2023-0071.html) (the
"Marvin Attack" timing side-channel that can leak RSA private-key material to a
network-observable attacker). As of `async-opcua` 0.14.0, `Cargo.lock` resolves
`rsa 0.9.10`, which is still inside the affected range.

## Impact

Any consumer who wants to use `SecurityPolicy::Basic256Sha256` (or the other
RSA-based policies) with a real X.509 client certificate is exposed to this
timing side channel during the secure channel handshake.

We hit this while integrating `async-opcua` into an industrial IoT edge
project. We currently hard-block every non-`SecurityPolicy::None` /
non-anonymous session in our adapter specifically because of this advisory
(tracked against `cargo-deny`/`cargo-audit` as a documented, reviewed
exception), which means we can only run unauthenticated, unencrypted OPC UA
sessions today — not something we're comfortable shipping to users who need
certificate-based auth.

## What we found researching this

- RustCrypto/RSA has an active but not-yet-complete migration to
  `crypto-bigint` for constant-time modular arithmetic (RustCrypto/RSA#390).
  Currently released versions still rely on blinding as a mitigation, and the
  advisory remains open with no fixed release yet.
- We couldn't find any existing issue or discussion here about this specific
  dependency chain, so we wanted to raise it directly rather than assume it's
  already on your radar.

## Question for maintainers

- Is there any appetite for making the crypto backend in `async-opcua-crypto`
  pluggable (e.g. an alternate implementation via `openssl` or `aws-lc-rs` for
  consumers who need the RSA-based security policies today), or is the plan
  to track/wait on `RustCrypto/rsa` landing a constant-time implementation?
- If a pluggable-backend approach seems reasonable, we'd be interested in
  contributing — wanted to check direction first before proposing anything
  larger.

Thanks for maintaining this crate — it's the most actively maintained
pure-Rust OPC UA implementation we've found.

EvanL1:
Follow-up: we spiked this locally to check feasibility before proposing anything, and wanted to share the result rather than just leave the question open.

## What we did

Forked `async-opcua-crypto` locally and swapped its use of the `rsa` crate for
[...]

PR #258, opened 2026-08-02
fix(crypto): replace rsa crate with openssl backend in async-opcua-crypto

EvanL1:
## Summary

This replaces `async-opcua-crypto`'s use of the RustCrypto `rsa` crate with the `openssl` crate, removing the dependency chain affected by [RUSTSEC-2023-0071](https://rustsec.org/advisories/RUSTSEC-2023-0071.html) (the "Marvin Attack" timing side-channel that can leak RSA private-key material to a network-observable attacker). Background and discussion in #257.

## What changed

`PrivateKey`, `PublicKey`, and `KeySize` in `aes/rsa_private_key.rs` are reimplemented on top of `openssl::pkey`/`openssl::rsa`/`openssl::sign`/`openssl::encrypt` instead of `rsa`/`pkcs1`/`pkcs8`/`pss`/`pkcs1v15`. The public API (type names, method names, method signatures) is unchanged, so nothing outside this crate needed to move. `policy/aes.rs`'s padding-scheme selection was adapted away from `rsa::traits::PaddingScheme`. `x509.rs` needed a small adapter type so the openssl-backed key can still be handed to `x509_cert::builder::CertificateBuilder`, which requires the signer to implement `signature::Keypair` + `spki::DynSignatureAlgorithmIdentifier` + `signature::Signer` — traits the `rsa` crate's `pkcs1v15::SigningKey` satisfied for free. `certificate_store.rs` needed a one-line PEM label swap. `rsa` is fully removed from `async-opcua-crypto`'s dependency tree; it's still used elsewhere in the workspace (`async-opcua-client`), which is out of scope here.

## Testing

`cargo test -p async-opcua-crypto` passes 30/30 with no changes to any test file, including every test that exercises the RSA-based security policies: `asymmetric_encrypt_and_decrypt` across all five policies (`Basic128Rsa15`, `Basic256`, `Basic256Sha256`, `Aes128Sha256RsaOaep`, `Aes256Sha256RsaPss`), `sign_verify_sha1`/`sign_verify_sha256`/`sign_verify_sha256_pss`, `calculate_cipher_text_size2`, `encrypt_decrypt_password`, and the cert-store PEM/DER round-trip tests. `cargo clippy -p async-opcua-crypto --all-targets` is clean. `cargo tree -p async-opcua-crypto -i rsa` no longer resolves.

## Things worth your attention during review

`to_der()` now round-trips through a PEM string internally, since `openssl::pkey::PKey::private_key_to_der()` only emits legacy PKCS#1 and there's no direct PKCS#8-DER export in the openssl crate. Functionally identical output, but it's an indirection I'd rather have called out than hidden. `PrivateKey::new`'s error type changes from `rsa::Error` to an opaque crate-local error — a public API signature change, though the single call site's behavior is unaffected. PSS salt length was confirmed to match (digest length, 32 bytes for SHA-256) by reading both implementations' defaults, not by cross-checking against a byte-for-byte fixture from another OPC UA stack, so an extra pair of eyes there would help.

## Open questions from #257, repeated here since they affect whether this is mergeable as-is

Is `openssl` an acceptable added dependency for this crate — it brings a C toolchain / cross-compile requirement that pure-Rust `rsa` didn't — or would you prefer something like `aws-lc-rs` instead? And is a pluggable backend something you want at all, versus just waiting on `RustCrypto/rsa`'s constant-time migration? Happy to rework this PR based on your answer either way; opened it now with the working version so there's something concrete to react to.


einarmo:
I commented on the issue. I don't think we're going to make this change at this time.

EvanL1:
Makes sense, thanks for the context — didn't mean to get in the way of the ECC work. Good to know a rsa fix is close too. We'll keep tracking this as a documented exception on our end and revisit once either lands. Closing this out.

PR #261, opened 2026-08-21
fix(client): skip server certificate parse for SecurityPolicy::None (#197)

### LINKED ISSUE #197: Client still gives BadCertificateInvalid when Security Policy is None and Server isn't sending any certificates
Image

vedantkulkarni:
## Summary

Fixes #197.

A client connecting with `SecurityPolicy::None` fails at `CreateSession` with `BadCertificateInvalid` when the server returns a **non-empty server certificate that fails strict DER parsing** — even though the certificate is never used under `None` (no signing, no encryption).

## Root cause

`Session::create_session` guards the server-certificate *validation* block with `if security_policy != SecurityPolicy::None`, but then calls `channel.update_from_created_session(...)` **unconditionally**:

```
update_from_created_session
  └─ SecureChannel::set_remote_cert_from_byte_string
       └─ X509::from_byte_string  ->  x509_cert::Certificate::from_der  (strict)
```

For a non-empty cert this always parses. `x509-cert` is intentionally stricter than OpenSSL, so real-world server certificates (in my case a PLC's OPC-UA server) that OpenSSL-based clients accept are rejected here, and the session fails — despite the certificate being irrelevant for `None`.

The failure is also hard to diagnose because `X509Error` is opaque (`From<x509::der::Error> for X509Error` discards the underlying `der::Error`).

## Fix

Skip parsing the remote certificate when the security policy is `None`, in `SecureChannel::set_remote_cert_from_byte_string`. This mirrors the existing empty-certificate skip added in #131 — in both cases the certificate is not needed, so there's no reason to parse (and reject) it.

```rust
self.remote_cert = if self.security_policy == SecurityPolicy::None
    || remote_cert.is_null_or_empty()
{
    None
} else {
    Some(X509::from_byte_string(remote_cert)?)
};
```

Nine-line change, no behavior change for secure policies (Sign / SignAndEncrypt still parse and validate the server certificate as before).

## Validation

Reproduced the original failure and confirmed the fix against a **live PLC OPC-UA server** (`SecurityPolicy::None`, anonymous) that ships a server certificate `x509-cert` rejects:

- **Before:** every `CreateSession` → `Failed to update channel from created session: BadCertificateInvalid: X509Error`, connection never establishes.
- **After:** session connects, and a batched `read()` of 562 nodes returns values normally end-to-end.

## Notes / possible follow-up

The opaque `X509Error` made this take longer to pin down than it should have. Surfacing the underlying `der::Error` (or its message) would help future debugging — happy to do that in a separate PR if desired.


einarmo:
The failure seems to be _mostly_ clippy and a deprecation from 1.98. I'll look into that. You will need to format the code though, `cargo fmt --all`.

I think it's fine to accept this, though I will note that strictly speaking a server that sends an invalid certificate, even when no certificate is required, is violating the standard.

emmanuel099:
Hi, thanks for fixing this bug.

I would suggest to fix it the client instead of the secure channel. The client already parses and validates the certificate or ignores it when security policy is none. The problem is, that the same certificate is parsed again in the channel in `update_from_created_session`. This can easily be solved by passing on the already parsed (optional) certificate. This keeps the secure channel as is (and `set_remote_cert_from_byte_string` keeps doing what its name suggests) and avoids duplicating the fallback behavior.

Rebased my patch https://github.com/emmanuel099/async-opcua/commit/5b885e40b14519dacbf1d3db7e86fb50474c4025 on top of master. This has been used in production since April 2026 without any problems, just forgot to upstream it.

> I think it's fine to accept this, though I will note that strictly speaking a server that sends an invalid certificate, even when no certificate is required, is violating the standard.

We recently had a problem with a recent update of IBA OPC-UA server that created non-standard compliant certificates with an overly long serial number., violating the lengths limit defined in RFC5280. Until such bugs are fixed, having the option to disable the security policy (and ignoring invalid certs) is great to have :)

Have a nice day!

einarmo:
Having looked at both solutions I think @emmanuel099's is more elegant, so we're going to use that one. Either way it should fix your problem, let me know if it doesn't. Thanks for the contribution!

I'll get it through soon enough, just need to clean up the breakage from the compiler version bump.

einarmo:
As mentioned, closed in favor of #263.

PR #263, opened 2026-08-24
fix(client): Properly handle server certificate on channel upgrade

### LINKED ISSUE #197: Client still gives BadCertificateInvalid when Security Policy is None and Server isn't sending any certificates

emmanuel099:
# Root Cause

The session already correctly ignored the server certificate when security policy is None. Unfortunately, the channel upgrade parsed the same certificate again, but without the fallback logic, meaning invalid certificates broke the client even when security policy was None.

# Fix

Instead of re-parsing the certificate, the session now passes the already-parsed and validated certificate to the channel (or None when security policy is None).

# Testing

We recently had a problem with an update of IBA OPC-UA server that created non-standard compliant certificates with an overly long serial number., violating the lengths limit defined in RFC5280. Setting security policy to None had no effect before this patch, the client failed with invalid certificate error. With this patch, the certificate is now properly ignored when security policy is None and connection to IBA OPC-UA is properly established.

Fixes #197


einarmo (approved):
Thanks!

PR #264, opened 2026-09-09
Accept a certificate chain in X509::from_der

vukoidra:
## What

`X509::from_der` fails when the DER it is given holds more than one certificate.

## Why it matters

OPC UA Part 6 §6.2.3 lets an `ApplicationInstanceCertificate` carry a chain: the certificate itself,
followed by the ones that issued it. A server whose certificate is signed by a CA rather than by
itself normally sends the whole chain.

`Certificate::from_der` treats everything after the first certificate as trailing data and rejects
the byte string. A client then fails to open a secure channel with
`BadCertificateInvalid: X509Error`, which reads like the server refusing the client and is not — the
client never got as far as asking.

Measured against Siemens WinCC Unified, which is signed by its own project CA: all ten endpoints it
advertises carry 2,266 bytes of certificate, exactly two concatenated DER certificates. Not one of
its endpoints can be used by this crate today. The .NET and Java stacks connect to the same server
without trouble.

## The change

Decode the leading certificate with a `SliceReader` and leave the rest. The certificate the protocol
is about is the first one; the remainder is the issuer chain, which this crate does not use.

This is the DER half of the `//keep certificate chain for another story` note already in `from_pem`.

## Test

`from_der_reads_the_first_certificate_of_a_chain` builds a two-certificate chain from the existing
2048-bit test helper and asserts the leading certificate comes back. It fails on `master` and passes
with the change. `cargo test -p async-opcua-crypto` is green: 31 passed.

## Not covered

The issuer chain is decoded past, not kept — validating a certificate against the CA it carries is
still out of scope, exactly as before. Nothing that used to parse parses differently.


vukoidra:
Reopened as #265 from our organisation's fork, so the branch does not sit on a personal account. Same two commits, same change.

